Issue #52, opened 2018-02-16
Document inter-build caching strategies


mattmoor:
This issue is intended to track documenting (and if necessary designing / implementing) facilities for inter-build caching.

PR #228, opened 2018-07-06
Add integration test demonstrating use of PersistentVolumeClaims for inter-build caching

ImJasonH:
Functional example of #52 

## Proposed Changes

  * Test defines a `PersistentVolumeClaim` which is attached first to a build that populates the cache, then another build reads the cache

**Release Note**

```release-note
NONE
```


mattmoor:
@ImJasonH you should sync with @srinivashegde86 who has been moving some of our common test code into `knative/pkg` to better enable this.

In the meantime, it might be interesting to write such a test in the `knative/serving` repo?

ImJasonH:
#341 instead

PR #294, opened 2018-08-02
Show additional build state info in CLI

### LINKED ISSUE #233: Need to return build status in UX CLI output
/area API
/kind dev

From a UX perspective `kubectl get builds` should return a status of sorts:

```
$ kubectl get builds
NAME          CREATED AT
hello-build   9m
```

Maybe a `COMPLETED` column with the number of completed steps.

greghaynes:
Add 'completed' column to 'get' cli to clearly indicate whether a build
has completed. Also add 'completed steps' field which can be used to
more easily determine current build status.

Fixes #233

**Release Note**

```release-note 
Add completed column and stepsCompleted to CLI
```

### DIFF
--- a/config/300-build.yaml
+++ b/config/300-build.yaml
@@ -25,3 +25,10 @@ spec:
     - all
     - knative
   scope: Namespaced
+  additionalPrinterColumns:
+  - name: Age
+    type: date
+    JSONPath: .metadata.creationTimestamp
+  - name: Completed
+    type: date
+    JSONPath: .status.completionTime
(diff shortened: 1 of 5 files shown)


ImJasonH:
It looks like this was added in Kubernetes 1.11, can we assume clients prior to that will just ignore this new information?

greghaynes:
Good question - I just tested out a ko apply using kubectl v1.9.9 and it succeeded while ignoring the additional columns when doing a get.

greghaynes:
I signed it!

mattmoor:
Related: https://github.com/knative/serving/issues/1138

PR #297, opened 2018-08-06
Update to 1.11 go-client library.

mattmoor:


mattmoor:
/hold

I need to figure out the right version of this `go-autorest` library to use now.

PR #309, opened 2018-08-23
Show intermediate update of build steps

### LINKED ISSUE #304: Update build with step states during build execution
## Expected Behavior

As a build executes on-cluster, `watch kubectl get build foo -oyaml` should show the current state of completed/ongoing steps.

## Actual Behavior

`.Spec.Status.StepStates` is only updated upon completion of the pod executing the build.

The on-cluster builder implementation uses [`registerDoneCallback`](https://github.com/knative/build/blob/50d7e70b0a13f380a498c109f945cb20e5acd84a/pkg/builder/cluster/builder.go#L218) to receive notifications about pod state when the pod completes, but it should also get updates during execution to update its `stepStates`.

  1. Create a build with multiple steps, each of which takes some amount of time (e.g., steps that each `sleep 10`)
  2. `kubectl apply -f the-build.yaml`
  3. `watch kubectl get build the-build -oyaml`
  4. The build's status is not updated until the build completes (successfully or otherwise)

shashwathi:
Fixes #304 

## Proposed Changes

  * Intermediate build steps like `waiting`, `running`.

### DIFF
--- a/test/multiple-steps/build-with-more-steps.yaml
+++ b/test/multiple-steps/build-with-more-steps.yaml
@@ -1,0 +1,29 @@
+# Copyright 2018 The Knative Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+apiVersion: build.knative.dev/v1alpha1
+kind: Build
+metadata:
+  name: test-build-with-more-steps
+  labels:
+    expect: succeeded
+spec:
+  steps:
+  - name: step1
+    image: ubuntu
+    command: ["/bin/bash"]
+    args: ["-c", "sleep 15"]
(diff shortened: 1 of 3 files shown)


test/multiple-steps/build-with-more-steps.yaml
```diff
@@ -0,0 +1,29 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+apiVersion: build.knative.dev/v1alpha1
+kind: Build
+metadata:
+  name: test-build-with-more-steps
```
ImJasonH: We have a few tests that happen to specify multiple steps, e.g., `test-custom-volume` and `test-home-volume`.

It's a real limitation of our integration testing "framework" that the only way we have right now to test things against a real cluster is to send builds and expect them to succeed or fail. Once we have a more "real" framework in place (stolen from `knative/serving` no doubt), we can have a test that runs a multi-step build and watches to see multiple status updates.

Until then, I'm not sure this test really adds much, except that it's proof that a build that has two steps succeeds.

ImJasonH:
The rest of this change looks **great** though, and I can't wait to get updates about the build's status while it's ongoing.

/lgtm
/approve
:tada:

test/multiple-steps/build-with-more-steps.yaml
shashwathi: I completely agree with you on this.

PR #310, opened 2018-08-23
Allow users to define cluster-scoped BuildTemplates #302

### LINKED ISSUE #302: Allow users to define cluster-scoped BuildTemplates
Some cluster operators might want to provide and manage predefined build templates, available across all namespaces.

We should create a `ClusterBuildTemplate` resource type which is available cluster-wide.

### Questions
* If a user creates a `BuildTemplate` in their namespace with a conflicting name with an existing `ClusterBuildTemplate` should we allow that or reject it?
* If we allow it, a Build would have to specify `template.name: 'some-namespace/the-template'` to specify the namespaced template.
* If we disallow it, creating a new `ClusterBuildTemplate` would have to look for a conflicting name across all namespaces.

I'm not familiar enough with Kubernetes API `ClusterFoo` idioms to know which is more consistent, hoping for pointers from those with more experience.

pivotal-nader-ziada:
Fixes #302

## Proposed Changes

  * Add new CRD for ClusterBuildTemplate
  * Update Build type to include Template Kind

**Release Note**
```
Add capability to define a BuildTemplate with a Cluster scope
```


pivotal-nader-ziada:
@cppforlife @dprotaso can you guys take a look? Thanks

pivotal-nader-ziada:
the integration test is failing with the following error
```
cannot list clusterbuildtemplates.build.knative.dev at the cluster scope: Unknown user "system:serviceaccount:knative-build:build-controller"
```
might be related to RBAC, looking into it

adrcunha:
Something is still wrong with your test, as the E2E test isn't able to fetch its status:

```
I0823 20:17:00.666] ERROR: test test-custom-source-cluster= but should be succeededtrue
I0823 20:17:00.821] ERROR: one or more tests failed
W0823 20:17:00.997] error: status is not found
```

Are you able to run it manually successfully?

pivotal-nader-ziada:
@adrcunha I wasn't getting an error locally when I made the first push, and I still don't get this error locally, but I get the following one 
```
mutation failed: clusterbuildtemplates.build.knative.dev is forbidden: User "system:serviceaccount:knative-build:build-controller" cannot list clusterbuildtemplates.build.knative.dev in the namespace "knative-build": Unknown user "system:serviceaccount:knative-build:build-controller"
```
which I though might be related to my cluster, but still investigating

mattmoor:
@pivotal-nader-ziada you need something like [this](https://github.com/knative/build/blob/5f0b563ce9064e49530b2e209b17b4d1bd71e99b/config/200-clusterrole.yaml#L18)

pivotal-nader-ziada:
@ImJasonH changes are completed now, can you take a look. Thanks

pkg/builder/interface.go
```diff
@@ -48,7 +48,7 @@ type Interface interface {
 	Builder() v1alpha1.BuildProvider
 
 	// Validate a Build for this flavor of builder.
-	Validate(*v1alpha1.Build, *v1alpha1.BuildTemplate) error
+	Validate(*v1alpha1.Build, v1alpha1.BuildTemplateInterface) error
```
ImJasonH: It looks like the template arg of this is never used, so we can probably remove it. Up to you whether you'd like to do that in this PR or I can do it after this is merged.

pkg/webhook/build.go
```diff
@@ -53,23 +53,29 @@ func (ac *AdmissionController) validateBuild(ctx context.Context, _ *[]jsonpatch
...
 		// Look up the template in the Build's namespace.
-		tmpl, err = ac.buildClient.BuildV1alpha1().BuildTemplates(b.Namespace).Get(tmplName, metav1.GetOptions{})
-		if err != nil {
-			return err
+		if b.Spec.Template.Kind == v1alpha1.ClusterBuildTemplateKind {
+			tmpl, err = ac.buildClient.BuildV1alpha1().ClusterBuildTemplates().Get(tmplName, metav1.GetOptions{})
+			if err != nil {
+				return err
+			}
+		} else {
```
ImJasonH: We should reject a build that specifies `kind: Nonsense` instead of assuming anything besides `ClusterBuildTemplate` means `BuildTemplate`.

Should we reject builds where `kind` is not specified? Or assume they mean one or the other?

pkg/webhook/cluster_build_template.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/knative/build/pkg/logging"
+)
+
+func (ac *AdmissionController) validateClusterBuildTemplate(ctx context.Context, _ *[]jsonpatch.JsonPatchOperation, old, new genericCRD) error {
+	_, tmpl, err := unmarshalClusterBuildTemplates(ctx, old, new)
+	if err != nil {
+		return err
+	}
+
+	if err := validateSteps(tmpl.Spec.Steps); err != nil {
```
ImJasonH: Can we share these between the two types of build templates?

test/custom-source/1-template.yaml
```diff
@@ -28,3 +28,4 @@ spec:
       ]
   template:
     name: dump-build-template
+    kind: BuildTemplate
```
ImJasonH: Since other tests specify a `template:` without explicitly specifying a `kind:`, it looks like this is unnecessary, and anything besides `kind: ClusterBuildTemplate` will be interpreted to mean `BuildTemplate`, including unspecified.

If we want an unspecified kind to default to `BuildTemplate`, I think that's okay but we should document it at least.

pkg/builder/interface.go
pivotal-nader-ziada: removed the unused template

config/300-clusterbuildtemplate.yaml
```diff
@@ -0,0 +1,34 @@
...
+    plural: clusterbuildtemplates
+    categories:
+    - all
+    - knative
+  scope: Cluster
+  additionalPrinterColumns:
+  - name: Age
+    type: date
+    JSONPath: .metadata.creationTimestamp
+  - name: Completed
```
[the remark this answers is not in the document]
pivotal-nader-ziada: removed from both

pkg/webhook/build.go
pivotal-nader-ziada: added a check to make sure the kind is either `BuildTemplate` or `ClusterBuildTemplate` and assumed `BuildTemplate` as the default if nothing is specified. Will also throw an error if other values are used.

pkg/webhook/cluster_build_template.go
pivotal-nader-ziada: added a template_common.go file with the common functions called by both types of templates

test/custom-source/1-template.yaml
pivotal-nader-ziada: Added a comment in the type definition and in the webhook, and will work on a PR for the `buildtemplates` repo to update the README explaining how to use Build Templates

PR #315, opened 2018-08-27
Add build timeout

### LINKED ISSUE #254: Add build timeout
## Expected Behavior
If a user configures build and corresponding k8s resources(schedule pod) do not get allocated under certain time then build controller updates the status as "FAILED" and reason as "Timed out waiting for resource allocation"

## Actual Behavior
If a user configures build and corresponding k8s resources(schedule pod) do not get allocated then build controller waits forever.

  1. Configure build with `node selector` that does not exist
  2. Check pod status. It should be in Pending state. 
  3. Run `kubectl get build` and build status should be "PENDING"

## Additional Info
- If build is marking the status as `FAILED` then it should clean up scheduling pod. 

/area API
/kind dev
/assign @shashwathi

## Questions
1. What is the default timeout ? 
1. Is this configurable by user? If so what should be the key in yaml?

shashwathi:
Fixes #254 

## Proposed Changes

  * Introduce build timeout under `timeout`
  * Default timeout is 10min.

### DIFF
--- a/pkg/apis/build/v1alpha1/build_types.go
+++ b/pkg/apis/build/v1alpha1/build_types.go
@@ -72,7 +72,7 @@ type BuildSpec struct {
 	// +optional
 	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
 
-	// Time after which the build times out. Defaults to 10minutes.
+	// Time after which the build times out. Defaults to 10 minutes.
 	//  Go's ParseDuration documentation for expected format: https://golang.org/pkg/time/#ParseDuration
 	Timeout string `json:"timeout,omitempty"`
 }
--- a/pkg/apis/build/v1alpha1/build_types.go
+++ b/pkg/apis/build/v1alpha1/build_types.go
@@ -72,8 +72,8 @@ type BuildSpec struct {
 	// +optional
 	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
 
-	// Time after which the build times out.
-	// Defaults to 10minutes.
+	// Time after which the build times out. Defaults to 10minutes.
+	//  Go's ParseDuration documentation for expected format: https://golang.org/pkg/time/#ParseDuration
 	Timeout string `json:"timeout,omitempty"`
 }
 
--- a/pkg/apis/build/v1alpha1/build_types.go
+++ b/pkg/apis/build/v1alpha1/build_types.go
@@ -34,8 +34,9 @@ type Build struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   BuildSpec   `json:"spec"`
-	Status BuildStatus `json:"status"`
+	Spec    BuildSpec   `json:"spec"`
+	Status  BuildStatus `json:"status"`
+	Timeout string      `json:"timeout,omitempty"`
 }
 
 // BuildSpec is the spec for a Build resource.
--- a/pkg/apis/build/v1alpha1/build_types.go
+++ b/pkg/apis/build/v1alpha1/build_types.go
@@ -34,9 +34,8 @@ type Build struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec    BuildSpec   `json:"spec"`
-	Status  BuildStatus `json:"status"`
-	Timeout string      `json:"timeout,omitempty"`
+	Spec   BuildSpec   `json:"spec"`
+	Status BuildStatus `json:"status"`
 }
 
 // BuildSpec is the spec for a Build resource.
@@ -72,6 +71,10 @@ type BuildSpec struct {
 	// More info: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/
 	// +optional
 	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+
+	// Time after which the build times out.
+	// Defaults to 10minutes.
+	Timeout string `json:"timeout,omitempty"`
 }
 
 // TemplateInstantiationSpec specifies how a BuildTemplate is instantiated into
--- a/pkg/builder/cluster/builder.go
+++ b/pkg/builder/cluster/builder.go
@@ -70,11 +70,7 @@ func (op *operation) Checkpoint(status *v1alpha1.BuildStatus) error {
 }
 
 func (op *operation) Terminate() error {
-	err := op.builder.kubeclient.CoreV1().Pods(op.namespace).Delete(op.name, &metav1.DeleteOptions{})
-	if err != nil {
-		return err
-	}
-	return nil
+	return op.builder.kubeclient.CoreV1().Pods(op.namespace).Delete(op.name, &metav1.DeleteOptions{})
 }
 
 func (op *operation) Wait() (*v1alpha1.BuildStatus, error) {
--- a/pkg/builder/cluster/builder.go
+++ b/pkg/builder/cluster/builder.go
@@ -69,6 +69,14 @@ func (op *operation) Checkpoint(status *v1alpha1.BuildStatus) error {
 	return nil
 }
 
+func (op *operation) Terminate() error {
+	err := op.builder.kubeclient.CoreV1().Pods(op.namespace).Delete(op.name, &metav1.DeleteOptions{})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
 func (op *operation) Wait() (*v1alpha1.BuildStatus, error) {
 	podCh := make(chan *corev1.Pod)
 	defer close(podCh)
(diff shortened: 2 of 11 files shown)


pkg/apis/build/v1alpha1/build_types.go
```diff
@@ -34,8 +34,9 @@ type Build struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   BuildSpec   `json:"spec"`
-	Status BuildStatus `json:"status"`
+	Spec    BuildSpec   `json:"spec"`
+	Status  BuildStatus `json:"status"`
+	Timeout string      `json:"timeout,omitempty"`
```
ImJasonH: I think we'll want `Timeout` to exist inside of `BuildSpec`

pkg/builder/cluster/builder.go
```diff
@@ -69,6 +69,14 @@ func (op *operation) Checkpoint(status *v1alpha1.BuildStatus) error {
 	return nil
 }
 
+func (op *operation) Terminate() error {
+	err := op.builder.kubeclient.CoreV1().Pods(op.namespace).Delete(op.name, &metav1.DeleteOptions{})
```
ImJasonH: nit: `return op.builder.kubeclient.CoreV1().Pods(op.namespace).Delete(op.name, &metav1.DeleteOptions{})`

pkg/controller/build/controller.go
```diff
@@ -274,8 +275,39 @@ func (c *Controller) syncHandler(key string) error {
 			if err != nil {
 				return err
 			}
-			if err := c.waitForOperationAsync(build, op); err != nil {
-				return err
+
+			// Check if build has timed out
```
ImJasonH: Will `syncHandler` be called periodically? Or only when there are updates to the build object? If it's periodic, how often is it called?

I worry that if this code is only run when a build updates (i.e., the pod's status changes), then a build that specifies one hour-long `sleep` step with a 10m timeout will not have any occasion to check for timeout until that step finishes, at which point the build has long exceeded its timeout.

pkg/controller/build/controller.go
```diff
@@ -274,8 +275,39 @@ func (c *Controller) syncHandler(key string) error {
...
+			if builder.IsTimeout(&build.Status, build.Timeout) {
+				//cleanup operation and update status
+				timeoutMsg := fmt.Sprintf("Build %q failed to finish within %q", build.Name, build.Timeout)
+
+				err = op.Terminate()
```
ImJasonH: In cases where the method only returns an err, I tend to prefer:

```go
if err := op.Terminate(); err != nil {
  c.logger.Error(...)
  return err
}
```

test/build-with-timeout/build-with-low-timeout.yaml
```diff
@@ -0,0 +1,26 @@
...
+  name: test-with-low-timeout
+  labels:
+    expect: failed
+timeout: 50s
+spec:
+  steps:
+  - name: step1
+    image: ubuntu
+    command: ["/bin/bash"]
+    args: ["-c", "sleep 2000"]
```
ImJasonH: Oh okay, so this test indicates that `syncHandler` must be called periodically, or at least not only when steps start/finish. Do we have any idea when or how often exactly `syncHandler` is called?

test/build-with-timeout/build-with-low-timeout.yaml
```diff
@@ -0,0 +1,26 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+apiVersion: build.knative.dev/v1alpha1
+kind: Build
+metadata:
+  name: test-with-low-timeout
+  labels:
+    expect: failed
```
ImJasonH: This is probably another case where our integration tests can be improved, but it would be great if we could have `expect: timeout` and have `e2e-test.sh` assert not only that the test failed, but that it failed due to timeout specifically. If the build was changed to `exit 1` the test would still pass. :confused: 

Anyway, not necessary for this PR, but something we should think about for future integration testing improvements.

pkg/apis/build/v1alpha1/build_types.go
shashwathi: That makes sense. 👍

test/build-with-timeout/build-with-low-timeout.yaml
shashwathi: Yeah `failed` is very generic and asserting on specific error would be ideal. I added controller level test for timeout which triggers `syncHandler` twice manually.

pkg/apis/build/v1alpha1/build_types.go
```diff
@@ -71,6 +71,10 @@ type BuildSpec struct {
 	// More info: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/
 	// +optional
 	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+
+	// Time after which the build times out.
+	// Defaults to 10minutes.
```
ImJasonH: typo: `10 minutes`.

This should maybe also link to Go's `ParseDuration` documentation to clarify the expected format: https://golang.org/pkg/time/#ParseDuration

Should we enforce a _maximum_ timeout too? Today, GCB limits builds to 24h for our own implementation purposes, though I suppose there's no strong reason to carry that limitation over to Knative Build. OTOH, it's easier to remove a restriction than to add one later. WDYT?

pkg/builder/interface.go
```diff
@@ -70,6 +75,25 @@ func IsDone(status *v1alpha1.BuildStatus) bool {
 	return false
 }
 
+// IsTimedOut returns true if the build's status indicates that build is timedout.
+func IsTimeout(status *v1alpha1.BuildStatus, buildTimeout string) bool {
+	defaultTimeout := "10m"
```
ImJasonH: It might be simpler to have the default already stated in terms of a `time.Duration`:

```go
var timeout time.Duration
if buildTimeout == "" {
  timeout = 10*time.Minute
} else {
  timeout, err = time.ParseDuration(buildTimeout)
  if err != nil {
    return err
  }
}
```

pkg/controller/build/controller.go
shashwathi: From logs it appears 30 sec is the sync interval

pkg/controller/build/controller.go
```diff
@@ -347,6 +379,15 @@ func (c *Controller) waitForOperationAsync(build *v1alpha1.Build, op builder.Ope
...
 
+func newBuildTimeoutError(message string) *errors.StatusError {
+	return &errors.StatusError{metav1.Status{
+		Status:  metav1.StatusFailure,
+		Code:    http.StatusGatewayTimeout,
+		Reason:  metav1.StatusReasonTimeout,
+		Message: message,
+	}}
+}
+
```
pivotal-nader-ziada: Are you using the http package  and error to indicate an timeout from an external server? Does the other status errors include an http error?

test/build-with-timeout/build-with-low-timeout.yaml
shashwathi: 30 seconds is the sync interval

pkg/builder/cluster/builder_test.go
```diff
@@ -130,6 +130,17 @@ func TestBasicFlow(t *testing.T) {
...
 		t.Fatal("timed out in op.Wait()")
 	})
+
+	// Trigger termination of pod
+	err = op.Terminate()
+	if err != nil {
+		t.Errorf("Expected no error while terminating operation")
+	}
+	// Verify pod is not available
+	_, err = podsclient.Get(op.Name(), metav1.GetOptions{})
```
ImJasonH: nit: This can be combined:

```go
if _, err := podsclient.Get(...); err == nil {
  t.Fatalf(...)
}
```

pkg/controller/build/controller.go
shashwathi: Status error package includes client timeouts but in this case I wanted to throw custom timeout error with more detail about build.

pkg/controller/build/controller.go
ImJasonH: Hmm, we should document that somewhere, that we effectively only enforce timeouts to a ~30s resolution. Maybe we should also `Watch` the pod, with a timeout? If we do that, do we need to worry about the controller job restarting while builds are ongoing, and losing state? (All questions for future work, not blocking this PR)

pkg/controller/build/controller.go
ImJasonH: I'm not actually sure what benefit using a `StatusError` gives us, can we just return `fmt.Errorf("Build timed out after %s", timeout)`?

shashwathi: Yeah I am good with that

shashwathi:
@ImJasonH 

> Hmm, we should document that somewhere, that we effectively only enforce timeouts to a ~30s resolution.

I will submit PR for docs. 

> Maybe we should also Watch the pod, with a timeout? If we do that, do we need to worry about the controller job restarting while builds are ongoing, and losing state?

I did consider the idea of `Watch`ing the Pod with timeout but that would require huge refactor and I wanted to keep the scope small. 
I can follow up with a POC PR to get some feedback on `Watch`. We can use callbacks map to track the Watch object for builds to avoid losing state. We need to be cautious of not resetting channel also.

Issue #179, opened 2018-06-06
git-init and creds-init images are too big


ImJasonH:
`git-init` and `creds-init` images are both based on `gcr.io/cloud-builders/gcloud:latest`, which includes every optional gcloud component, and clocks in at ~2 GB. This makes builds slow, and `ko apply -L -f config/` slow because the image is always rebuilt and pushed into the daemon (https://github.com/google/go-containerregistry/issues/205)

`git-init` probably only needs `git`, and `creds-init` probably only needs `ssh-keygen`, so base images that include those packages (or a single base image that includes both) would help a lot.

pivotal-nader-ziada:
All the tests ran successfully with a base image based on alpine. The `init-base` image using alpine is only 23.2 MB vs 112.3 MB for image using ubuntu 

```
FROM alpine:latest
RUN apk add --update git openssh-client
```
To help with running the builds faster on a smaller image, I have the following proposal

### Proposal

1. Maintain this `init-base` image separately from the release process to allow for updates based on its base image (FROM entry) to be published as needed.
2. Use automated builds linked to the Dockerfile to publish the `init-base` image on a publicly available repo, for example `gcr.io/knative-build/init-base-image`
3. As long as the `.ko.yaml` is pulling latest, it would pick it the latest image irrespective of when it was updated, and we can use tags in case we need to make sure some release is using a specific version of the image. 
4. Since this is only one Dockerfile, we probably don’t need to have this in a separate repo similar to [cloud-sdk-docker](https://github.com/GoogleCloudPlatform/cloud-sdk-docker) or [cloud-builders](https://github.com/GoogleCloudPlatform/cloud-builders), but this is also an option.

@mattmoor @ImJasonH thoughts?

ImJasonH:
That seems totally reasonable to me. I don't have a strong preference about `alpine`, we can always change it later if we decide we want to.

PR #317, opened 2018-08-27
Dockerfile for new minimal base image based on alpine

pivotal-nader-ziada:
Related to https://github.com/knative/build/issues/179

## Proposed Changes

  * Add a Dockerfile to generate a minimal base image to be used by build
  * Once this image is published, the `.ko.yaml` file should be updated to point to it

**Release Note**
```
Base image used by build now based on alpine to make downloads faster
```

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -1,0 +1,4 @@
+FROM alpine:latest
+  
+RUN apk add --update git openssh-client
+


images/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM alpine:latest
+  
+RUN apk add --update git openssh-client
```
ImJasonH: Should this run `apk update` too?

pivotal-nader-ziada: I believe the --update flag fetches the current package index before adding the package.

PR #322, opened 2018-08-29
Add pod start time

### LINKED ISSUE #321: Surface pod start time in build status
/area API
/kind dev

/assign @shashwathi

## Expected Behavior
Build status should include pos start time in status. This will help developer get more information about pod latency and startup times.

## Actual Behavior
No pod start time in status

shashwathi:
Fixes #321

## Proposed Changes

  * Rename current status.StartTime to status.CreationTime
  * Add pod start time as status.StartTime


shashwathi:
I **really** want to add this build example as part of this PR as it shows `startTime` set to null and `creationTime` set to build created time.

Ideally the following example should fail because pods are never created and build times out after 3 minutes. The problem is that as the build times out, pods are terminated and [pod termination is unsuccessful](https://github.com/knative/build/blob/master/pkg/controller/build/controller.go#L286
) (because pod does not exist) and build status is not updated to failed which causes build to remain in `Unknown` state.


```
apiVersion: build.knative.dev/v1alpha1
kind: Build
metadata:
  name: test-build-with-badselector
  labels:
    expect: failed
spec:
  timeout: 3m
  steps:
  - name: step1
    image: ubuntu
    command: ["/bin/bash"]
    args: ["-c", "sleep 15"]
  nodeSelector:
    disk: "fake-ssd" 
```

One way to approach this problem is instead of just bubbling of all pod termination errors, it might be better to filter out `Not Found` errors. Another easy approach would be to update build status before calling `Terminate()` function(not a fan of this idea). 

thoughts ? ideas?
@ImJasonH @pivotal-nader-ziada @dprotaso

pkg/builder/nop/builder_test.go
```diff
@@ -45,8 +45,11 @@ func TestBasicFlow(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Unexpected error executing OperationFromStatus: %v", err)
 	}
+	if bs.CreationTime.IsZero() {
+		t.Errorf("bs.CreationTime; want non-zero, got %v", bs.CreationTime)
+	}
 	if bs.StartTime.IsZero() {
-		t.Errorf("bs.StartTime; want non-zero, got %v", bs.StartTime)
+		t.Errorf("status.StartTime; want non-zero, got %v", bs.StartTime)
```
ImJasonH: Can you update the other error messages for consistency? Or return this to `bs.`

pkg/controller/build/controller_test.go
```diff
@@ -232,12 +232,13 @@ func TestTimeoutFlows(t *testing.T) {
...
 	build := newBuild("test")
-	addBuffer, err := time.ParseDuration("10m")
+	buffer, err := time.ParseDuration("10m")
 	if err != nil {
 		t.Errorf("Error parsing duration")
 	}
 
-	build.Status.StartTime.Time = metav1.Now().Time.Add(-addBuffer)
+	// Set creation time to Now()-10min
+	build.Status.CreationTime.Time = metav1.Now().Time.Add(buffer)
```
ImJasonH: Previously this was `-addBuffer`, now it's `buffer`, can you explain why this is?

shashwathi: It made more sense to call the variable as buffer. Ideally I want to subtract the buffer time to trigger timeout.

shashwathi: Also just realized that line is not required for test setup.  Let me update.

ImJasonH:
I think we should filter out not found errors from `Terminate`, which I imagine could happen for other reasons too.

shashwathi:
@ImJasonH : I have addressed your comments

pkg/builder/cluster/builder.go
```diff
@@ -21,16 +21,16 @@ import (
 	"fmt"
 	"sync"
 
+	v1alpha1 "github.com/knative/build/pkg/apis/build/v1alpha1"
```
ImJasonH: Can we separate these first-party imports (anything from `knative/build`) from the vendored imports from other packages?

pkg/builder/interface.go
```diff
@@ -92,8 +92,7 @@ func IsTimeout(status *v1alpha1.BuildStatus, buildTimeout string) bool {
 			return false
 		}
 	}
-
-	return time.Since(status.StartTime.Time).Seconds() > timeout.Seconds()
+	return time.Since(status.CreationTime.Time).Seconds() > timeout.Seconds()
```
ImJasonH: Time spent scheduling the pod shouldn't count toward the build's timeout, should this consider `status.StartTime` instead of `status.CreationTime` ?

Unless we were thinking this would be a useful way to kill a build that was unschedulable, in which case we might need to introduce two separate timeouts.

In GCB we only enforce timeouts once the build starts, since it's not the user's fault if it takes us a long time to start their build. We've considered enforcing a timeout on queued time, but I don't think any user has asked for it.

test/columns.txt
```diff
@@ -1,2 +1,2 @@
-NAME	BUILDER	TYPE	STATUS	START	END
-.metadata.name	.status.builder	.status.conditions[*].state	.status.conditions[*].status	.status.startTime	.status.completionTime
+NAME	BUILDER	TYPE	STATUS	CREATED  START	END
```
ImJasonH: Is it possible to add columns to show startup duration and build execution duration? I don't know how powerful this template language is, if it's hard then don't worry about it, we can try to add it later.

Also, we can remove the `BUILDER` columns now that there's only one builder implementation.

pkg/builder/interface.go
shashwathi: If we don't enforce timeout for allocating resources then build will remain in `pending` state forever. I was planning to follow up with a PR to introduce another timeout for allocating resources(queue timeout). Probably default this timeout to 10min and user doesn't need to worry about using this.

test/columns.txt
shashwathi: If the build has started but pod hasn't been scheduled then is it okay to have negative duration? Might be better off calculating that in status if we need that info.

ImJasonH: We should avoid a negative duration, since that would be pretty hard to interpret. I'm fine putting this off for some future change where we add startup and build duration to the `status` and make it easy to display here. Don't worry about it for now.

pkg/builder/interface.go
ImJasonH: Okay, if that's on your radar for future changes that sounds good to me.

shashwathi: Created: https://github.com/knative/build/issues/325 to follow up with changes discussed in this PR.

PR #329, opened 2018-09-05
 Add simple e2e tests in Go for no-op success and failure builds

ImJasonH:
This uses some test infrastructure from `knative/pkg`, namely flags and kubeclient setup.

It uses a `TestMain` to ensure that the test namespace `build-tests` exists in the specified cluster (in the process ensuring the cluster exists and is reachable), then runs two builds, one that succeeds and one that fails.

Future improvements should eventually replace all the YAML-based tests in `test/`, and more, since with this extra level of control we can do things like check for logs, interact with the build's underlying pod (e.g., what happens when the underlying pod dies?), check timing (e.g., build only ran during its configured timeout), check state transitions (e.g., build was `Pending`, then running, then successful), can kill a build while it's running or pending, and can help us track performance metrics over time.

**Release Note**
```release-note
NONE
```

/assign @shashwathi 
/assign @mattmoor


test/e2e-tests.sh
```diff
@@ -66,9 +66,11 @@ kubectl delete buildtemplates --all
 # Run the tests
 
 header "Running tests"
-
 ko apply -R -f test/ || fail_test
 
+header "Running Go e2e tests"
+GOCACHE=off go test -tags e2e ./test/e2e/... || fail_test
```
adrcunha: Use report_go_test here instead of `go test` so you'll get individual results on testgrid and in the gubernator report.

Also, don't `fail_test` here. Fail after checking the results, so you'll get all results of the test. Better run the go tests first, since you're not checking the results here (even better: move everything to a function). Something like:

```
local failed=0
report_go_test ... || failed=1
run_old_ugly_tests_in_shell_script || failed=1
ko apply ... || failed=1
(( failed )) && abort_test
success
```

test/e2e/README.md
```diff
@@ -0,0 +1,25 @@
+# End-to-End testing
+
+## Running the tests
+
+To run all e2e tests:
+```
+GOCACHE=off go test -tags e2e ./test/e2e/...
```
adrcunha: This info seems to conflict with using `count=1` as per https://golang.org/doc/go1.10#test

test/e2e/simple_test.go
```diff
@@ -0,0 +1,122 @@
...
+func TestMain(m *testing.M) {
+	flag.Parse()
+	clients, err := newClients(test.Flags.Kubeconfig, test.Flags.Cluster, buildTestNamespace)
+	if err != nil {
+		log.Fatalf("newClients: %v", err)
+	}
+
+	// Ensure the test namespace exists, by trying to create it and ignoring
+	// already-exists errors.
+	if _, err := clients.kubeClient.Kube.CoreV1().Namespaces().Create(&corev1.Namespace{
```
adrcunha: This is a nice thing, we should adopt it for `serving` as well, instead of having to create a test space as a requirement.

PR #331, opened 2018-09-05
Useless check err

### LINKED ISSUE #330: Useless check err return
kind of refactor

/assign @zrss

`waitForOperationAsync` will always return nil

https://github.com/knative/build/blob/4dfa5dabd59de26ef5607b02ac7a536cfb05ecea/pkg/controller/build/controller.go#L379-L392

it seems there is no need to check err here

https://github.com/knative/build/blob/4dfa5dabd59de26ef5607b02ac7a536cfb05ecea/pkg/controller/build/controller.go#L307-L312

zrss:
Fixes #330 

## Proposed Changes

  * change waitForOperationAsync to waitForOperation

**Release Note**

```release-note
NONE
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,6 @@
 
 # VSCode config
 .vscode
+
+# JetBrains IDE config
+.idea
--- a/pkg/controller/build/controller.go
+++ b/pkg/controller/build/controller.go
@@ -305,12 +305,10 @@ func (c *Controller) syncHandler(key string) error {
 
 				c.logger.Errorf("Timeout: %v", timeoutMsg)
 				return nil
-			} else {
-				// if not timed out then wait async
-				if err := c.waitForOperationAsync(build, op); err != nil {
-					return err
-				}
 			}
+
+			// if not timed out then wait async
+			go c.waitForOperation(build, op)
 		} else {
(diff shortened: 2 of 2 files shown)


shashwathi:
My thoughts on this PR. I agree that `waitForOperationAsync` function was always returning nil so no need for error check in controller. However the changes in this PR include changing function to be synchronous function but the caller still does not handle error.

I would consider updating build controller to not include error check(for `waitForOperationAsync` function) and update function definition to not return error.

```
func (c *Controller) waitForOperationAsync(build *v1alpha1.Build, op builder.Operation) {
   go func(){
   }()
return
} 
```

(error not returned ^)

```
 c.waitForOperationAsync(build, op) 
```

(no error handling ^)

I think this makes the contract more readable. Function is explicitly stating its async and its not returning channels for events/errors so caller does not need to worry about error handling.

shashwathi:
I agree that both options are not great. Probably we need a new approach altogether to handle this path.  
I will defer to @ImJasonH @mattmoor @adrcunha @dprotaso for expert opinions

zrss:
thanks shashwathi, ImJasonH advice

/test pull-knative-build-build-tests
/test pull-knative-build-integration-tests
/test pull-knative-build-unit-tests

PR #335, opened 2018-09-05
Update build execution time

shashwathi:
## Proposed Changes

  * Update build execution time to be (build finish time - pod start time)
 * Add integration test to check build status


test/e2e/simple_test.go
```diff
@@ -120,3 +120,63 @@ func TestFailingBuild(t *testing.T) {
...
+		t.Fatalf("watchBuild did not return expected BuildTimeout error")
+	}
+
+	// verify reason for build failure is timeout
+	if b.Status.GetCondition(v1alpha1.BuildSucceeded).Reason != "BuildTimeout" {
+		t.Fatalf("wanted BuildTimeout; got %q", b.Status.GetCondition(v1alpha1.BuildSucceeded).Reason)
+	}
+}
+
+func TestBuildWithHighTimeout(t *testing.T) {
```
pivotal-nader-ziada: Not sure you need this test here, since the build is actually successful

test/e2e/simple_test.go
```diff
@@ -120,3 +120,63 @@ func TestFailingBuild(t *testing.T) {
...
+	}); err != nil {
+		t.Fatalf("Error creating build: %v", err)
+	}
+
+	b, err := clients.buildClient.watchBuild(buildName)
+	if err == nil {
+		t.Fatalf("watchBuild did not return expected BuildTimeout error")
+	}
+
+	// verify reason for build failure is timeout
```
ImJasonH: Can we add a check that the build executed for 20s, +/- some wiggle room? That wiggle room will be pretty high today, since we only poll every 30s.

test/e2e/simple_test.go
```diff
@@ -120,3 +120,63 @@ func TestFailingBuild(t *testing.T) {
...
+				Args:    []string{"-c", "sleep 2000"},
+			}},
+		},
+	}); err != nil {
+		t.Fatalf("Error creating build: %v", err)
+	}
+
+	b, err := clients.buildClient.watchBuild(buildName)
+	if err == nil {
+		t.Fatalf("watchBuild did not return expected BuildTimeout error")
```
ImJasonH: This could probably be `t.Error`, and we can keep checking other things about the returned build.

test/e2e/simple_test.go
ImJasonH: +1, our test that a build under the timeout passes is `TestSimpleBuild`

shashwathi: I agree. I wanted to test a build with higher timeout set. As an alternate idea I could alter the `SimpleBuild` test to have a timeout as well.

test/e2e/simple_test.go
shashwathi: Okay how about we check if build duration is between buildTimeout and buildTimeout + 40s(30s poll + 10s tolerance)?

ImJasonH: Yeah that sounds good. Add a TODO to reduce the wiggle room once we're more strict, and maybe run the test a few times to make sure it's not flaky.

shashwathi: I ran into couple of edge cases while re-running the test. So I fixed couple of bugs in this PR.

shashwathi:
@ImJasonH @pivotal-nader-ziada : Addressed all comments and fixed e2e tests. Please take a look at this and provide feedback

pkg/builder/interface.go
```diff
@@ -75,7 +75,8 @@ func IsDone(status *v1alpha1.BuildStatus) bool {
 	return false
 }
 
-// IsTimedOut returns true if the build's status indicates that build is timedout.
+// IsTimeOut returns true if the build's execution time is greater than
```
mattmoor: nit: `IsTimeout` (lowercase `o`)

test/e2e/simple_test.go
```diff
@@ -82,6 +83,7 @@ func TestSimpleBuild(t *testing.T) {
 			Name:      buildName,
 		},
 		Spec: v1alpha1.BuildSpec{
+			Timeout: "40s",
```
mattmoor: Can we use an actual `time.Duration` for this field?

shashwathi: Thats a great time. I have updated the PR with this change.

pkg/apis/build/v1alpha1/build_types.go
```diff
@@ -74,7 +75,8 @@ type BuildSpec struct {
 
 	// Time after which the build times out. Defaults to 10 minutes.
 	//  Go's ParseDuration documentation for expected format: https://golang.org/pkg/time/#ParseDuration
-	Timeout string `json:"timeout,omitempty"`
+	//Timeout string `json:"timeout,omitempty"`
```
ImJasonH: Delete this commented-out line, and now you can probably also remove the line about Go's ParseDuration format.

pkg/webhook/build.go
```diff
@@ -237,15 +237,14 @@ func validationError(reason, format string, fmtArgs ...interface{}) error {
...
 		return nil
 	} else {
-		timeout, err := time.ParseDuration(timeout)
-		if err != nil {
-			return validationError("InvalidTimeFormat", "invalid build timeout %q err %#v. Refer https://golang.org/pkg/time/#ParseDuration for time format documentation", timeout, err)
+		if timeout < 0 {
+			return validationError("InvalidTimeFormat", "invalid build timeout %q. Refer https://golang.org/pkg/time/#ParseDuration for time format documentation", timeout)
 		}
 
 		// time out should not greater than 24 hours
```
ImJasonH: GitHub apparently won't let me comment on lines that weren't changed in this PR :unamused: but I think we don't need `Hours` on the line below this. Just `timeout > maxTimeout` has the same effect.

shashwathi:
If timeout is not specified then it will be equal to 0 and thats valid input. so I will update the logic to below


```
if timeout < 0 {
  // bad
} else if timeout > maxTimeout {
  // bad
}
// good
```

shashwathi:
Go serializes Duration in nanoseconds. User needs to be aware of this conversion. I have added more comments in spec and test as well.

pivotal-nader-ziada:
I would prefer as a user to enter the build timeout in seconds, not nanoseconds.

shashwathi:
I will update the PR with build duration as string. 👍

shashwathi:
@ImJasonH  : Addressed all comments. Let me know if you have any comments on this PR.

test/build-with-timeout/build-with-low-timeout.yaml
```diff
@@ -18,7 +18,7 @@ metadata:
   labels:
     expect: failed
 spec:
-  timeout: 50s
+  timeout: "50s"
```
ImJasonH: These quotes aren't necessary right? I don't feel strongly about removing them, I just want to make sure this all works if a user doesn't explicitly quote the string.

test/e2e/simple_test.go
```diff
@@ -120,3 +122,90 @@ func TestFailingBuild(t *testing.T) {
 		t.Fatalf("watchBuild did not return expected error: %v", err)
 	}
 }
+
+func TestBuildLowTimeout(t *testing.T) {
+	clients := setup(t)
+
+	buildName := "build-low-timeout"
+	buildTimeout := "50s"
```
ImJasonH: Optional: `buildTimeout = 50*time.Second` then `Timeout: buildTimeout.String()` means you don't have to parse and handle the error down on line 169. Up to you, it's also fine like this.

test/e2e/simple_test.go
```diff
@@ -120,3 +122,90 @@ func TestFailingBuild(t *testing.T) {
...
+	// verify reason for build failure is timeout
+	if successCondition.Reason != "BuildTimeout" {
+		t.Fatalf("wanted BuildTimeout; got %q", successCondition.Reason)
+	}
+	buildDuration := b.Status.CompletionTime.Time.Sub(b.Status.StartTime.Time).Seconds()
+	lowerEnd, err := time.ParseDuration(buildTimeout)
+	if err != nil {
+		t.Errorf("Error parsing build duration: %v", err)
+	}
+	higherEnd := 90 * time.Second // build timeout + 30 sec poll time + 10 sec
```
ImJasonH: If this value is intended to be based on `buildTimeout`, maybe it should explicitly be based on it?

`higherEnd := buildTimeout + 30*time.Second + 10*time.Second`

That way if we change `buildTimeout` above we won't also have to remember to update this value.

test/build-with-timeout/build-with-low-timeout.yaml
shashwathi: Quotes are not necessary. When I changed the type into string I might have gone on a spree to add quotes everywhere :D I will follow up with a PR to address all your comments.

tcnghia:
@shashwathi @ImJasonH  can we use https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/duration.go instead of strings here?

shashwathi:
Thats a great idea. We rejected the idea of using `time.Duration` because native Golang time Duration does not support strings unmarshal. I can follow up with a PR to use k8s duration type. Thanks for finding that @tcnghia

PR #336, opened 2018-09-05
Wrap YAML E2E tests in a function

adrcunha:
This makes the code cleaner, and allow failing faster on error.

### DIFF
--- a/test/e2e-tests.sh
+++ b/test/e2e-tests.sh
@@ -45,9 +45,11 @@ function abort_test() {
 
 function run_yaml_tests() {
   header "Running YAML E2E tests"
+  echo ">> Startting tests"
   ko apply -R -f test/ || return 1
 
   # Wait for tests to finish.
+  echo ">> Waiting for tests to finish"
   local tests_finished=0
   for i in {1..60}; do
     local finished="$(kubectl get builds --output=jsonpath='{.items[*].status.conditions[*].status}')"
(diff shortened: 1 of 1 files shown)


test/e2e-tests.sh
```diff
@@ -65,47 +109,12 @@ kubectl delete buildtemplates --all
...
 
-local failed=0
+failed=0
+
 header "Running Go e2e tests"
 report_go_test -tags e2e ./test/e2e/... -count=1 || failed=1
 
-header "Running YAML e2e tests"
-ko apply -R -f test/ || failed=1
+run_yaml_tests || failed=1
```
ImJasonH: Should this be `&& failed=1` ?

If `run_yaml_tests` returns `0`, that'll be interpreted as `false` and therefore trip the `||` branch.

adrcunha: No. In UNIX world, 0 means success.

ImJasonH: Right. I knew as I was typing it there was no way I had found a real bug. 😄

ImJasonH:
It looks like my recent e2e test change broke the YAML tests, and it wasn't this change. Either way, the YAML tests are effectively skipped. Do you happen to see where my bug is?

adrcunha:
The YAML tests don't seem to be skipped. If you think they are because gubernator only shows "2 tests passed" (the go tests you added), it's because it doesn't parse the output of the shell tests. Unless we intend to keep these tests forever the way they are (instead of rewriting them as go tests), I don't think it's worth updating the script to parse their results.

adrcunha:
Only now I saw your initial comment explaining why you think the tests are being skipped. I don't think they are, because there's a 20s interval between the builds being created and the banner "all tests passed":

```
I0905 22:22:31.546] build "test-workspace-volume" created
I0905 22:22:52.847] **************************************
I0905 22:22:52.847] ***        ALL TESTS PASSED        ***
I0905 22:22:52.847] **************************************
```

However, this is indeed poor logging. Let me fix that in this PR.

PR #339, opened 2018-09-07
Move build pod information into build status, remove builder cruft

ImJasonH:
This removes some more cruft left over from the Google-specific builder implementation which was removed in #305 

**Release Note**
```release-note
Breaking API change: .Status.Cluster.PodName is now .Status.PodName, and there is no more .Status.Builder.
```


ImJasonH:
Seems like we can't guarantee this breaking change is safe unless we rev the CRD version, which is also annoying. Closing this for now.

PR #340, opened 2018-09-07
Trim StepStates for implicit steps

### LINKED ISSUE #338: Filter out StepStates for implicit steps
## Expected Behavior

When I run a build that has N steps, I see N `StepStates`, one for each step I requested.

## Actual Behavior

I see N+1 `StepStates`, or N+2 if my build specified source and had to be fetched.

When populating `StepStates`, we should filter out the first one, which is the build running `creds-init`, and if the build specifies source, we should filter out the second one as well.

If we add other implicit steps (e.g., build output collection), we should make sure to filter those out too.

ImJasonH:
Fixes #338 

## Proposed Changes

  * If a build specifies no source, skip only the first `StepState`
  * If the build specifies source, skip the first _two_ `StepState`s
  * Add tests, and move `diffmatchpatch` helper to `pkg/buildtest`

**Release Note**
```release-note
.Status.StepStates omits states for implicit steps, so the number of StepStates == number of user-specified build steps
```

/assign @shashwathi

### DIFF
--- a/pkg/builder/cluster/builder.go
+++ b/pkg/builder/cluster/builder.go
@@ -46,8 +46,7 @@ func (op *operation) Name() string {
 	return op.name
 }
 
-func (op *operation) Checkpoint(build *v1alpha1.Build) error {
-	status := build.Status
+func (op *operation) Checkpoint(build *v1alpha1.Build, status *v1alpha1.BuildStatus) error {
 	status.Builder = v1alpha1.ClusterBuildProvider
 	if status.Cluster == nil {
 		status.Cluster = &v1alpha1.ClusterSpec{}
@@ -65,6 +64,9 @@ func (op *operation) Checkpoint(build *v1alpha1.Build) error {
 		// If the build specifies source, skip another container status, which
 		// is the source-fetching container.
 		skip++
+	}
+	if skip > len(op.statuses) {
+		skip = 0
 	}
 
 	for _, s := range op.statuses[skip:] {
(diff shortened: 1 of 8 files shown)


pkg/builder/cluster/builder.go
```diff
@@ -57,7 +57,19 @@ func (op *operation) Checkpoint(status *v1alpha1.BuildStatus) error {
...
-	for _, s := range op.statuses {
+
+	// Always ignore the first pod status, which is creds-init.
+	skip := 1
+	if build.Spec.Source != nil {
+		// If the build specifies source, skip another container status, which
+		// is the source-fetching container.
+		skip++
+	}
+	if skip > len(op.statuses) {
```
shashwathi: Looks like there is no test coverage for this condition `skip> len(op.statuses)`.

mattmoor:
I wonder if we are losing potentially valuable signal when credential initialization fails?  Creds has proven to be a source of friction with the Build CRD, and so I just want to make sure we're not making things worse.

ImJasonH:
> I wonder if we are losing potentially valuable signal when credential initialization fails? Creds has proven to be a source of friction with the Build CRD, and so I just want to make sure we're not making things worse.

That's a reasonable concern. I think the solution there is to put some more work into surfacing creds-init and source-fetching statuses into the build status itself, probably via termination messages.

These implicit steps should be a transparent implementation detail to the user, and they shouldn't need to know how many implicit steps we add at the beginning (and possibly end) of their specified steps, to be able to tell which of their steps may have failed.

shashwathi:
> That's a reasonable concern. I think the solution there is to put some more work into surfacing creds-init and source-fetching statuses into the build status itself, probably via termination messages.

I like this idea. We can loop through the `Init containers` and only surface the errors if there are any. Ignore implicit steps if they are not providing value to the user.

PR #347, opened 2018-09-13
Add skeleton reconcilers for the core resources.

mattmoor:
This also hooks them up to ./cmd/controller and drops the template controllers, which weren't doing any more than this is.


pkg/reconciler/build/build.go
```diff
@@ -0,0 +1,136 @@
...
+
+	// Convert the namespace/name string into a distinct namespace and name
+	namespace, name, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		logger.Errorf("invalid resource key: %s", key)
+		return nil
+	}
+
+	// Get the Build resource with this namespace/name
+	_, err = c.buildsLister.Builds(namespace).Get(name)
```
ImJasonH: nit:

```go
if _, err := c.buildsLister...; errors.IsNotFound(err) {
  // not found 
} else if err != nil {
  return err
}
```

pkg/reconciler/buildtemplate/buildtemplate.go
```diff
@@ -0,0 +1,118 @@
...
+	_, err = c.buildTemplatesLister.BuildTemplates(namespace).Get(name)
+	if errors.IsNotFound(err) {
+		// The BuildTemplate resource may no longer exist, in which case we stop processing.
+		logger.Errorf("buildtemplate %q in work queue no longer exists", key)
+		return nil
+	} else if err != nil {
+		return err
+	}
+
+	// TODO: Meaningful reconciliation.
```
ImJasonH: What does this mean in this context? Is there any meaningful reconciliation to do for build templates? (same with `ClusterBuildTemplate`s)

mattmoor:
Hmm, I marked this WIP in the hopes of creating test skeletons for each of these, but we still haven't shared the [core testing bit](https://github.com/knative/serving/blob/master/pkg/reconciler/testing/table.go) in `knative/pkg` and copying it here feels gross if we're going to let them drift.

@dprotaso I believe you still have plans to improve this further?

pkg/reconciler/build/build.go
mattmoor: Done, although when we start acting on the other part we'll probably rehoist it.

ImJasonH:
/lgtm
/approve

Tests will be added in a future PR once knative/pkg gains the test helpers.

PR #349, opened 2018-09-13
Update knative/pkg to pull in the kmeta stuff and add assertions.

mattmoor:


mattmoor:
Just trying to break off more manageable pieces for review :)

PR #354, opened 2018-09-14
[WIP] Have the ClusterBuildTemplate controller instantiate Image resources.

mattmoor:
Unlike the BuildTemplate controller, which colocates the Image references, this places the Image resources (which are namespaced) in the knative-build namespace.

WIP until https://github.com/knative/build/pull/353 lands.


pkg/reconciler/buildtemplate/buildtemplate.go
```diff
@@ -114,14 +128,68 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
+		}
+
+		// Wait for the creates to complete.
+		if err := grp.Wait(); err != nil {
+			return err
+		}
+	}
+
+	// Delete any Image caches relevant to older versions of this resource.
+	propPolicy := metav1.DeletePropagationForeground
```
ImJasonH: Can you add a test that cached images are deleted upon template update?

pkg/reconciler/buildtemplate/buildtemplate.go
```diff
@@ -114,14 +128,68 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
+	eics, err := c.imagesLister.Images(bt.Namespace).List(kmeta.MakeVersionLabelSelector(bt))
+	if err != nil {
+		return err
+	}
+
+	// Make sure we have all of the desired caching resources.
+	if missing := allCached(ics, eics); len(missing) > 0 {
+		grp, _ := errgroup.WithContext(ctx)
+
+		for _, m := range missing {
```
ImJasonH: Can you add a test that new images are cached upon template update?

pkg/reconciler/clusterbuildtemplate/clusterbuildtemplate.go
```diff
@@ -107,14 +122,63 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
 
-	// TODO: Meaningful reconciliation.
+	if err := c.reconcileImageCaches(ctx, cbt); err != nil {
+		return err
+	}
+
 	return nil
 }
+
+func (c *Reconciler) reconcileImageCaches(ctx context.Context, cbt *v1alpha1.ClusterBuildTemplate) error {
```
ImJasonH: Can this be moved to a common location shared by both types of template reconcilers?

pkg/reconciler/clusterbuildtemplate/resources/imagecache_test.go
```diff
@@ -0,0 +1,231 @@
...
+	"github.com/google/go-cmp/cmp"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/knative/build/pkg/apis/build/v1alpha1"
+	"github.com/knative/build/pkg/system"
+	caching "github.com/knative/caching/pkg/apis/caching/v1alpha1"
+)
+
+func TestMakeImageCache(t *testing.T) {
```
ImJasonH: Duplicating this test also feels Wrong, if the code was in a single common location then you'd only need to have one test for it.

mattmoor: This checks that we properly create things in the system namespace.  I can reduce the testing, but the vast majority of the logic is shared :)

pkg/reconciler/clusterbuildtemplate/clusterbuildtemplate.go
mattmoor: I've made the parts that are common usable from the `buildtemplate` reconciler.

pkg/reconciler/buildtemplate/buildtemplate.go
```diff
@@ -193,3 +182,18 @@ func allCached(desired []caching.Image, observed []*caching.Image) (missing []ca
 	}
 	return missing
 }
+
+func CreateMissingImageCaches(ctx context.Context, client cachingclientset.Interface, missing []caching.Image) error {
```
ImJasonH: Could this just transparently calculate `missingImageCaches` and create them? Users shouldn't need to call both methods.

pkg/reconciler/clusterbuildtemplate/clusterbuildtemplate.go
```diff
@@ -107,14 +121,41 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
+	ics := resources.MakeImageCaches(cbt)
+
+	eics, err := c.imagesLister.Images(system.Namespace).List(kmeta.MakeVersionLabelSelector(cbt))
+	if err != nil {
+		return err
+	}
+
+	// Make sure we have all of the desired caching resources.
+	if missing := buildtemplate.MissingImageCaches(ics, eics); len(missing) > 0 {
```
ImJasonH: Taking this even further, `buildtemplate.ReconcileImageCaches` could:

1. Calculate missing images
1. Cache missing images
1. Uncache old images

pkg/reconciler/clusterbuildtemplate/resources/imagecache_test.go
ImJasonH: Yeah I think what I'm asking for is a test that does all of these cases, for either BuildTemplate or ClusterBuildTemplate, and checks that the resources are created either in the template's namespace or in the case of ClusterBuildTemplate, the system namespace. Sort of like a two-level nested table-driven test, does that make sense?

pkg/reconciler/clusterbuildtemplate/clusterbuildtemplate.go
mattmoor: I think I like the split we reach above since the creates can be done in a generic way without the namespace coming into play.  There's already a large number of arguments, and I generally prefer largely single-purpose functions.

pkg/reconciler/clusterbuildtemplate/resources/imagecache_test.go
mattmoor: Yes, let me spend some time on this.  Will ping this when RFAL.

mattmoor: ok this should be RFAL.

PR #358, opened 2018-09-14
Update knative/pkg to get kmeta changes

jonjohnsonjr:
Rename NewOwnerReference -> NewControllerRef.


adrcunha:
@jonjohnsonjr how did you update the dependencies? I just realized that a bunch of unnecessary stuff from knative/test-infra was added (`update-deps.sh` should get rid of them). I'll write a PR for cleaning it up and make sure the script is WAI.

PR #360, opened 2018-09-17
Test refactor

### LINKED ISSUE #350: Build e2e cannot run twice on an existing cluster
I see:
```
--- FAIL: TestSimpleBuild (0.09s)
    simple_test.go:93: Error creating build: builds.build.knative.dev "simple-build" already exists
```

I think we just want to delete the `build-tests` namespace in a trap within `./test/e2e-test.sh`?

shashwathi:
Fixes #350 

## Proposed Changes

  * Use `pkg/logger`
  * Remove namespace when e2e test exits
  * Remove build at end of each test


test/e2e/simple_test.go
```diff
@@ -38,45 +37,30 @@ import (
...
-		log.Printf("Namespace %q already exists", buildTestNamespace)
-	} else {
-		log.Fatalf("Error creating namespace %q: %v", buildTestNamespace, err)
-	}
+	clients := setup(logger)
+	test.CleanupOnInterrupt(func() { teardownNamespace(clients, logger) }, logger)
 
+	code := m.Run()
 	defer func() {
+		// Cleanup namespace
```
ImJasonH: It doesn't add much to call `defer` at the end of a function, since there's nothing to defer from. If `Run` panics, we won't yet know to defer teardown. I think this should maybe be:

```
...
defer teardownNamespace(clients, logger)
os.Exit(m.Run())
```


```

File: test/e2e/e2e.go [hunk 2]
```diff
@@ -42,12 +43,50 @@ type clients struct {
...
+		if err := clients.kubeClient.Kube.CoreV1().Namespaces().Delete(buildTestNamespace, &metav1.DeleteOptions{}); err != nil && !kuberrors.IsNotFound(err) {
+			logger.Fatalf("Error deleting namespace %q: %v", buildTestNamespace, err)
+		}
+	}
+}
+
+func teardownBuild(clients *clients, logger *logging.BaseLogger, name string) {
+	if clients != nil && clients.buildClient != nil {
+		logger.Infof("Deleting build %q in namespace %q", name, buildTestNamespace)
+		err := clients.buildClient.deleteBuild(name)
```
ImJasonH: nit: `if err := ...; err != nil {`

test/e2e/simple_test.go
shashwathi: If `os.Exit` is invoked then `defer` is not respected.
Ref: https://gobyexample.com/exit

shashwathi: In 2nd option I need to declare `code` var before defer so I will try 1st option without `defer`.

pivotal-nader-ziada: Actually thinking about it more, the first option will not `teardownNamespaces` if `m.Run` errors out, but if you have the defer before, it will be cleaned up

shashwathi: `m.Run` should return non zero exit code if it errors but if `m.Run` has any non recoverable errors then defer might good way to catch it all.

PR #362, opened 2018-09-18
Start to use the common condition infrastructure from knative/pkg.

mattmoor:
WIP until https://github.com/knative/build/pull/361 merges.

cc @n3wscott


pkg/apis/build/v1alpha1/build_types.go
```diff
@@ -209,12 +214,17 @@ type BuildStatus struct {
...
 	// StepStates describes the state of each build step container.
 	StepStates []corev1.ContainerState `json:"stepStates,omitEmpty"`
-	// Conditions describes the set of conditions of this build.
-	Conditions []BuildCondition `json:"conditions,omitempty"`
+
 	// StepsCompleted lists the name of build steps completed.
 	StepsCompleted []string `json:"stepsCompleted"`
+
+	// Conditions describes the set of conditions of this build.
+	Conditions duckv1alpha1.Conditions `json:"conditions,omitempty"`
```
n3wscott: you can update this to the following:

```
Conditions duck.Conditions `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
```

pkg/apis/build/v1alpha1/build_types.go
```diff
@@ -271,42 +260,28 @@ type BuildList struct {
...
 // GetCondition returns the Condition matching the given type.
-func (bs *BuildStatus) GetCondition(t BuildConditionType) *BuildCondition {
-	for _, cond := range bs.Conditions {
-		if cond.Type == t {
-			return &cond
-		}
-	}
-	return nil
+func (bs *BuildStatus) GetCondition(t duckv1alpha1.ConditionType) *duckv1alpha1.Condition {
+	return buildCondSet.Manage(bs).GetCondition(t)
```
n3wscott: LOL `manage bs`

pkg/apis/build/v1alpha1/build_types.go
```diff
@@ -271,42 +260,28 @@ type BuildList struct {
...
 }
 
 // SetCondition sets the condition, unsetting previous conditions with the same
 // type as necessary.
-func (b *BuildStatus) SetCondition(newCond *BuildCondition) {
-	if newCond == nil {
-		return
+func (bs *BuildStatus) SetCondition(newCond *duckv1alpha1.Condition) {
```
n3wscott: you should be able to remove this entirely and replace it with `Mark*` methods. Looks like there is only Succeeded and Failed. I would implement:

`MarkBuildSucceded`
`MarkBuildFailed`

and then remove SetCondition entirely.

pkg/builder/cluster/builder.go
```diff
@@ -75,7 +76,7 @@ func (op *operation) Checkpoint(build *v1alpha1.Build, status *v1alpha1.BuildSta
 		}
 		status.StepStates = append(status.StepStates, s.State)
 	}
-	status.SetCondition(&v1alpha1.BuildCondition{
```
n3wscott: this should be `status.Initialize()`

pkg/builder/cluster/builder.go
```diff
@@ -134,21 +135,21 @@ func (op *operation) Wait() (*v1alpha1.BuildStatus, error) {
 
 	if pod.Status.Phase == corev1.PodFailed {
 		msg := getFailureMessage(pod)
-		bs.SetCondition(&v1alpha1.BuildCondition{
```
n3wscott: And then use the correct `Mark` method

pkg/apis/build/v1alpha1/build_types.go
mattmoor: Ack.  Build hasn't yet adopted this model, so I think that should be a follow up.

mattmoor:
/hold

There is something going on here that I need to look into.

mattmoor:
/hold cancel

I had to update the e2e script since apparently the serialized field name changed.

pkg/apis/build/v1alpha1/build_types.go
mattmoor: This is only going to be meaningful for API servers.

PR #365, opened 2018-09-19
Use k8s Duration instead of string for build timeout

shashwathi:
Fixes #335 

## Proposed Changes

  * Follow up PR to update type from string to k8s Duration type

cc @tcnghia

### DIFF
--- a/pkg/webhook/build.go
+++ b/pkg/webhook/build.go
@@ -243,7 +243,7 @@ func validateTimeout(timeout metav1.Duration) error {
 	if timeout.Duration > maxTimeout {
 		return validationError("InvalidTimeout", "build timeout exceeded 24h")
 	} else if timeout.Duration < 0 {
-		return validationError("InvalidTimeFormat", "build timeout should be greater than 0")
+		return validationError("InvalidFormat", "build timeout should be greater than 0")
 	}
 	return nil
 }
--- a/pkg/webhook/build.go
+++ b/pkg/webhook/build.go
@@ -243,7 +243,7 @@ func validateTimeout(timeout metav1.Duration) error {
 	if timeout.Duration > maxTimeout {
 		return validationError("InvalidTimeout", "build timeout exceeded 24h")
 	} else if timeout.Duration < 0 {
-		return validationError("InvalidTimeFormat", "invalid build timeout %q. Refer https://golang.org/pkg/time/#ParseDuration for time format documentation", timeout)
+		return validationError("InvalidTimeFormat", "build timeout should be greater than 0")
 	}
 	return nil
 }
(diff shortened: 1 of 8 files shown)


pkg/webhook/build.go
```diff
@@ -237,19 +237,13 @@ func validationError(reason, format string, fmtArgs ...interface{}) error {
...
-		if err != nil {
-			return validationError("InvalidTimeFormat", "invalid build timeout %q err %#v. Refer https://golang.org/pkg/time/#ParseDuration for time format documentation", timeout, err)
-		}
-		if timeoutDuration > maxTimeout {
-			return validationError("InvalidTimeout", "build timeout exceeded 24h")
-		}
+	if timeout.Duration > maxTimeout {
+		return validationError("InvalidTimeout", "build timeout exceeded 24h")
+	} else if timeout.Duration < 0 {
+		return validationError("InvalidTimeFormat", "invalid build timeout %q. Refer https://golang.org/pkg/time/#ParseDuration for time format documentation", timeout)
```
ImJasonH: This type should handle parse validation errors now, so this error message should just be "cannot define negative timeout" or something along those lines.

pkg/webhook/webhook_test.go
```diff
@@ -210,18 +211,18 @@ func TestValidateBuild(t *testing.T) {
 		reason: "invalid build timeout",
 		build: &v1alpha1.Build{
 			Spec: v1alpha1.BuildSpec{
-				Timeout: "garbagetimeout",
+				Timeout: metav1.Duration{Duration: -48 * time.Hour},
```
ImJasonH: This test case should be "negative timeout" now

pkg/webhook/build.go
```diff
@@ -237,19 +237,13 @@ func validationError(reason, format string, fmtArgs ...interface{}) error {
...
-		if err != nil {
-			return validationError("InvalidTimeFormat", "invalid build timeout %q err %#v. Refer https://golang.org/pkg/time/#ParseDuration for time format documentation", timeout, err)
-		}
-		if timeoutDuration > maxTimeout {
-			return validationError("InvalidTimeout", "build timeout exceeded 24h")
-		}
+	if timeout.Duration > maxTimeout {
+		return validationError("InvalidTimeout", "build timeout exceeded 24h")
+	} else if timeout.Duration < 0 {
+		return validationError("InvalidTimeFormat", "build timeout should be greater than 0")
```
ImJasonH: The reason should be `InvalidFormat` just like a timeout >24h, otherwise everything LGTM

pivotal-nader-ziada:
good straight forward change 
/lgtm

PR #366, opened 2018-09-19
Use errgroup

shashwathi:
## Proposed Changes

  * Use sync wait group in build controller

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -19,6 +19,7 @@ package main
 import (
 	"flag"
 	"log"
+	"sync"
 	"time"
 
 	"go.uber.org/zap"
@@ -134,10 +135,12 @@ func main() {
 			logger.Fatalf("failed to wait for cache at index %v to sync", i)
 		}
 	}
-
+	var wg sync.WaitGroup
 	// Start all of the controllers.
 	for _, ctrlr := range controllers {
+		wg.Add(1)
 		go func(ctrlr controller.Interface) {
+			defer wg.Done()
 			// We don't expect this to return until stop is called,
 			// but if it does, propagate it back.
 			if err := ctrlr.Run(threadsPerController, stopCh); err != nil {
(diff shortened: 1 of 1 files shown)


cmd/controller/main.go
```diff
@@ -134,18 +135,19 @@ func main() {
 			logger.Fatalf("failed to wait for cache at index %v to sync", i)
 		}
 	}
-
+	var wg sync.WaitGroup
```
ImJasonH: We could use https://godoc.org/golang.org/x/sync/errgroup here too, and just have each `g.Go` call `ctrlr.Run` and return its error.

```
var g errgroup.Group
for _, c := range controllers {
  c := c // capture c for the goroutine
  g.Go(func() error {
    return c.Run(...)
  })
}
if err := g.Wait(); err != nil {
  logger.Fatalf("...", err)
}
```

shashwathi: I have never used `errgroup` before. That looks great

cmd/controller/main.go
```diff
@@ -135,19 +135,20 @@ func main() {
...
-			if err := ctrlr.Run(threadsPerController, stopCh); err != nil {
-				logger.Fatalf("Error running controller: %s", err.Error())
-			}
-		}(ctrlr)
+			return ctrlr.Run(threadsPerController, stopCh)
+		})
+	}
+
+	// Wait for all controllers to finish and log errors if there are any
+	if err := g.Wait(); err == nil {
```
ImJasonH: `err != nil`

cmd/controller/main.go
```diff
@@ -134,18 +135,20 @@ func main() {
 			logger.Fatalf("failed to wait for cache at index %v to sync", i)
 		}
 	}
+	var g errgroup.Group
 
 	// Start all of the controllers.
 	for _, ctrlr := range controllers {
-		go func(ctrlr controller.Interface) {
+		ctrlr := ctrlr
```
pivotal-nader-ziada: what does this line do?

pivotal-nader-ziada: oh got it, maybe if the name was different, it would not be confusing

PR #367, opened 2018-09-19
Add test for build types

shashwathi:
## Proposed Changes

  * Increase test coverage in build types

### DIFF
--- a/pkg/apis/build/v1alpha1/build_template_types_test.go
+++ b/pkg/apis/build/v1alpha1/build_template_types_test.go
@@ -42,8 +42,8 @@ func TestBuildTemplateSpec(t *testing.T) {
 func TestBuildTemplateGroupVersionKind(t *testing.T) {
 	c := BuildTemplate{}
 
-	epectedKind := "BuildTemplate"
-	if c.GetGroupVersionKind().Kind != epectedKind {
-		t.Errorf("GetGroupVersionKind mismatch; expected: %v got: %v", epectedKind, c.GetGroupVersionKind().Kind)
+	expectedKind := "BuildTemplate"
+	if c.GetGroupVersionKind().Kind != expectedKind {
+		t.Errorf("GetGroupVersionKind mismatch; expected: %v got: %v", expectedKind, c.GetGroupVersionKind().Kind)
 	}
 }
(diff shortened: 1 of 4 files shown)


pkg/apis/build/v1alpha1/build_template_types_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+	if a := cmp.Diff(c.TemplateSpec(), expectedBuildSpec); a != "" {
+		t.Errorf("templateSpec mismatch; expected: %v got: %v", expectedBuildSpec, a)
+	}
+}
+
+func TestBuildTemplateGroupVersionKind(t *testing.T) {
+	c := BuildTemplate{}
+
+	epectedKind := "BuildTemplate"
```
ImJasonH: typo: `expectedKind`

pkg/apis/build/v1alpha1/build_types_test.go
```diff
@@ -57,16 +59,57 @@ func TestBuildConditions(t *testing.T) {
...
-		t.Fatalf("Unexpected Condition length; want 1, got %d", len(rev.Status.Conditions))
+	if cmp.Diff(foobuildCondition.Status, foo.Status) != "" {
+		t.Fatalf("Unexpected build condition status; want %v got %v", foo.Status, foobuildCondition.Type)
 	}
 
 	// Add a second condition.
-	rev.Status.SetCondition(bar)
+	b.Status.SetCondition(bar)
+
+	if len(b.Status.Conditions) != 2 {
```
ImJasonH: Can you `cmp.Diff` against the full expected conditions?

```
got := b.Status.Conditions()
want := []Condition{{
 // first
}, {
 // second
}}
if d := cmp.Diff(got, want); d != "" {
  t.Fatalf("got diff: %s", d)
}
```

(Printing the diff string is useful because it will include only exactly what changed, which can save a lot of time when debugging)

shashwathi: I did try to compare `Condition` object directly but there is timestamp which causes `cmp.Diff` to fail. Thats why I went down the path of adding assertions for `Status` and `Type` individually

shashwathi:
@ImJasonH : Addressed your comments

PR #373, opened 2018-09-24
[WIP] We now mount in emptyDir volumes for "$HOME/.ssh" and "$HOME/.docker"

### LINKED ISSUE #117: Consider exposing credentials via an environment variable that isn't $HOME
In the creds-init docs under the section [Exposing the credential integration](https://github.com/elafros/build/blob/master/cmd/creds-init/README.md#exposing-the-credential-integration) it says

> To solve this, prior to even the Source step, we will run a credential initialization process that accesses each of its secrets and aggregates them into their respective files in `$HOME`.

## Expected Behaviour

The `$HOME` env var exposed matches the pathname of the user's home directory. As per the POSIX definition.

## Actual Behaviour

Certain applications fetch the user's home directory using the syscalls [`getpw*`](http://man7.org/linux/man-pages/man3/getpwnam.3.html). This returns a different directory than  the`$HOME` env var that's set by the build controller for each step.

Write a go program and use the [os/user](https://golang.org/pkg/os/user) package to print the current user's home dir using `HomeDir`.

Make sure the program is compiled with cgo enabled. Enabling cgo uses the `getpwuid_r`

## Potential Solutions
- Update the entry in`/etc/passwd` have have `pw_dir` match the `cred-init` path
- Change the `cred-init` `$HOME` to match what's in the `/etc/passwd`
- **(preferred)** Use a different env var to refer to the location of the `cred-init` output

dprotaso:
This should resolve #117 as we don't need to change the "$HOME" env
var anymore

### DIFF
--- a/cmd/git-init/main.go
+++ b/cmd/git-init/main.go
@@ -18,7 +18,6 @@ package main
 import (
 	"bytes"
 	"flag"
-	"os"
 	"os/exec"
 
 	"github.com/knative/pkg/logging"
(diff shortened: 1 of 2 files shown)


ImJasonH:
Another benefit of having a persistent `$HOME` dir is to support tools that write config to `$HOME`, e.g., `git config user.name` or `gcloud config set project`. We got a fair number of confused users in GCB before we persisted `$HOME` across steps, since they would set some config in step 0 and it would be gone by step 1.

dprotaso:
@ImJasonH We could just mount an `emptyDir` to `/root`

ImJasonH:
@dprotaso Yeah, that's more or less the reason we put things in `/builder/`, since that was a namespace we thought we could reasonably lay claim to.

dprotaso:
closing as this does not address the case when a build step's user is non-root

PR #375, opened 2018-09-26
Move the use of `VerifyType` in tests

vdemeester:
Related to  knative/pkg#97

## Proposed Changes

Those calls to `duck.VerifyType` are done at runtime and thus could be
costly at program startup. Putting them under tests ensure we still
assert those types but during unit testing.


shashwathi:
@vdemeester : Thanks for the PR.

shashwathi:
Removing hold after reading slack conversation.

/hold cancel

PR #376, opened 2018-09-26
Controller test

### LINKED ISSUE #265: Bring up code coverage of controller to 80%
/area test-and-release
/kind doc

Expected Behavior
github.com/knative/build/pkg/controller code cover should be 80+%

Actual Behavior
It's at 50.5%.

Steps to Reproduce the Problem
Pls refer to https://testgrid-dot-knative-tests.appspot.com/knative-build#coverage for up-to-date CC numbers

Additional Info

shashwathi:
Fixes #265 

## Proposed Changes

  * Add tests for build controller
  * Update error msg when cluster build template/ build template is not found 
  * Add hooks in nop builder for inducing error scenarios 

/assign @ImJasonH 
/assign @pivotal-nader-ziada

### DIFF
--- a/pkg/builder/nop/builder.go
+++ b/pkg/builder/nop/builder.go
@@ -58,6 +58,9 @@ func (nb *operation) Checkpoint(_ *v1alpha1.Build, status *v1alpha1.BuildStatus)
 }
 
 func (nb *operation) Terminate() error {
+	if nb.builder.OpErr != nil {
+		return nb.builder.OpErr
+	}
 	return nil
 }
 
@@ -109,6 +112,9 @@ type Builder struct {
 
 	// Err is the error that should be returned from calls to this builder.
 	Err error
+
+	// Operation error
+	OpErr error
 }
 
 func (nb *Builder) Builder() v1alpha1.BuildProvider {
@@ -130,5 +136,8 @@ func (nb *Builder) BuildFromSpec(*v1alpha1.Build) (buildercommon.Build, error) {
 
 // OperationFromStatus returns the no-op operation.
 func (nb *Builder) OperationFromStatus(*v1alpha1.BuildStatus) (buildercommon.Operation, error) {
+	if nb.Err != nil {
+		return nil, nb.Err
+	}
 	return &operation{builder: nb}, nil
 }
--- a/pkg/builder/nop/builder.go
+++ b/pkg/builder/nop/builder.go
@@ -58,10 +58,7 @@ func (nb *operation) Checkpoint(_ *v1alpha1.Build, status *v1alpha1.BuildStatus)
 }
 
 func (nb *operation) Terminate() error {
-	if nb.builder.OpErr != nil {
-		return nb.builder.OpErr
-	}
-	return nil
+	return nb.builder.OpErr
 }
 
 func (nb *operation) Wait() (*v1alpha1.BuildStatus, error) {
--- a/pkg/builder/nop/builder_test.go
+++ b/pkg/builder/nop/builder_test.go
@@ -17,6 +17,8 @@ limitations under the License.
 package nop
 
 import (
+	"errors"
+
 	v1alpha1 "github.com/knative/build/pkg/apis/build/v1alpha1"
 	buildercommon "github.com/knative/build/pkg/builder"
 
@@ -129,5 +131,22 @@ func TestOperationFromStatus(t *testing.T) {
 	}
 	if msg, failed := buildercommon.ErrorMessage(status); failed {
 		t.Errorf("ErrorMessage(%v); wanted not failed, got %q", status, msg)
+	}
+}
+
+func TestTerminate(t *testing.T) {
+	builder := Builder{
+		OpErr: errors.New("testerr"),
+	}
+	op, err := builder.OperationFromStatus(&v1alpha1.BuildStatus{
+		Google: &v1alpha1.GoogleSpec{
+			Operation: operationName,
+		},
+	})
+	if err != nil {
+		t.Fatalf("Unexpected error waiting for builder.Operation: %v", err)
+	}
+	if op.Terminate() == nil {
+		t.Errorf("Expected error from operation.Terminate")
 	}
 }
--- a/pkg/controller/build/controller_test.go
+++ b/pkg/controller/build/controller_test.go
@@ -533,7 +533,7 @@ func TestTimeoutFlowWithFailedOperation(t *testing.T) {
 	}
 
 	// Update status to past time by substracting buffer time
-	first.Status.CreationTime.Time = metav1.Now().Time.Add(-buffer)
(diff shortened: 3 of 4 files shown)


pkg/builder/nop/builder.go
```diff
@@ -59,6 +59,9 @@ func (nb *operation) Checkpoint(_ *v1alpha1.Build, status *v1alpha1.BuildStatus)
 }
 
 func (nb *operation) Terminate() error {
+	if nb.builder.OpErr != nil {
```
ImJasonH: You can just return `nb.builder.OpErr` here.

pkg/controller/build/controller_test.go
```diff
@@ -299,3 +523,99 @@ func TestTimeoutFlows(t *testing.T) {
...
+	}
+
+	// Update status to past time by substracting buffer time
+	first.Status.CreationTime.Time = metav1.Now().Time.Add(-buffer)
+
+	// We have to manually update the index, or the controller won't see the update.
+	f.updateIndex(i, []*v1alpha1.Build{first})
+
+	// Run a second iteration of the syncHandler to receive error from operation.
+	if err := c.syncHandler(getKey(build, t)); err == nil || cmp.Diff(err.Error(), oppErr.Error()) != "" {
```
ImJasonH: This could just check `err != opErr`

shashwathi:
/hold

There is data race. Let me fix that

shashwathi:
@ImJasonH : Addressed your comments

pkg/builder/nop/builder_test.go
```diff
@@ -137,3 +139,20 @@ func TestOperationFromStatus(t *testing.T) {
 		t.Errorf("ErrorMessage(%v); wanted not failed, got %q", status, msg)
 	}
 }
+
+func TestTerminate(t *testing.T) {
+	builder := Builder{
+		OpErr: errors.New("testerr"),
+	}
+	op, err := builder.OperationFromStatus(&v1alpha1.BuildStatus{
+		Google: &v1alpha1.GoogleSpec{
```
pivotal-nader-ziada: Might be out of the scope of this PR, but we should rename GoogleSpec since we don't have google builder anymore

shashwathi: @ImJasonH mentioned in earlier comment that about cleaning this up in v1alpha2

ImJasonH: The only thing that uses `GoogleSpec` anymore is the nop builder, and I think as an aspiration I'd like it'd be cool to see if we can get rid of nop builder too, and just use a cluster builder implementation backed by fake k8s clients.

But yeah, as for dropping `GoogleSpec` I think we have to wait until v1alpha2 to make breaking changes, and CRD versioning won't have a good story for a couple more k8s releases it looks like :(

pkg/controller/build/controller_test.go
```diff
@@ -18,18 +18,27 @@ package build
 
 import (
 	"errors"
+	"fmt"
 	"strings"
 	"testing"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
```
ImJasonH: nit: I try to keep imports somewhat tidy, in just three blocks: go stdlib imports, non-`knative/build` imports, and then `knative/build` imports. The second block includes packages from `knative/pkg`, since those are vendored and aren't under our direct control in this repo.

pkg/controller/build/controller_test.go
```diff
@@ -95,12 +100,203 @@ func (f *fixture) updateIndex(i informers.SharedInformerFactory, bl []*v1alpha1.
...
+	defer close(stopCh)
+	defer close(eventCh)
+
+	c, i, k8sI := f.newController(bldr, eventCh)
+	f.updateIndex(i, []*v1alpha1.Build{build})
+	i.Start(stopCh)
+	k8sI.Start(stopCh)
+
+	// induce failure when fetching build information in controller
+	reactor := func(action clientgotesting.Action) (handled bool, ret runtime.Object, err error) {
```
ImJasonH: Ooooh fancy, I like this a lot :+1:

pkg/controller/build/controller_test.go
```diff
@@ -95,12 +100,203 @@ func (f *fixture) updateIndex(i informers.SharedInformerFactory, bl []*v1alpha1.
...
+	k8sI.Start(stopCh)
+
+	if err := c.syncHandler(getKey(build, t)); err != nil {
+		t.Errorf("Unexpected error while syncing build: %s", err.Error())
+	}
+}
+
+func TestBuildWithNonExistentTemplates(t *testing.T) {
+	tests := []struct {
+		bldr  builder.Interface
```
ImJasonH: `bldr` is only passed as `&nop.Builder{}`, so it doesn't need to be a field in the test case struct.

In fact, the only thing that varies between the two `build`s is the `Kind`, so maybe just make that the test case:

```
for _, kind := []string{BuildTemplateKind, ClusterBuildTemplateKind} {
  ...
}
```

pkg/controller/build/controller_test.go
```diff
@@ -95,12 +100,203 @@ func (f *fixture) updateIndex(i informers.SharedInformerFactory, bl []*v1alpha1.
...
+		eventCh := make(chan string, 1024)
+		defer close(stopCh)
+		defer close(eventCh)
+
+		c, i, k8sI := f.newController(test.bldr, eventCh)
+		f.updateIndex(i, []*v1alpha1.Build{test.build()})
+		i.Start(stopCh)
+		k8sI.Start(stopCh)
+
+		err := c.syncHandler(getKey(test.build(), t))
```
ImJasonH: nit: `if err := ...; err == nil { ... } else if !kuberrors.IsNotFound(err) { ... }`

pkg/controller/build/controller_test.go
```diff
@@ -95,12 +100,203 @@ func (f *fixture) updateIndex(i informers.SharedInformerFactory, bl []*v1alpha1.
...
+		if err == nil {
+			t.Errorf("Expect error syncing build")
+		}
+		if !kuberrors.IsNotFound(err) {
+			t.Errorf("Expect error to be not found err: %s", err.Error())
+		}
+	}
+}
+
+func TestBuilWithTemplate(t *testing.T) {
```
ImJasonH: typo: `TestBuildWithTemplate`

pkg/controller/build/controller_test.go
shashwathi: Okay I like that approach. I think on Vscode Go plugin re-arranges the imports so I have to manually check it didn't override my arrangement 👍

pkg/controller/build/controller_test.go
shashwathi: I will consider writing a git commit hook to catch this in future :D

Issue #378, opened 2018-09-26
Use pkg webhook controller in Build


shashwathi:
Build currently uses its own implementation of webhook controller. Update webhook to use pkg/wehbook. 
- pkg/webhook has better test coverage
- Avoid code duplication

/area API
/kind dev

/assign @shashwathi

PR #379, opened 2018-09-27
Use pkg/webhook along with build webhook implementation

shashwathi:
Address #378 

## Proposed Changes
  * Do not use custom build webhook implementation
  * Implement validate and default methods for all build CRDs and use pkg/wehbook controller.

**FYI**
 I have not added any tests in this PR. If this change does not get 👎 then I will add tests to this PR and also delete custom webhook code.

Goal of this PR is to get conversation started about this change

Lot of validation in build CRDs included calling k8s API to validate object exist like
secrets, build templates. I have not included that change here because I noticed that Serving did not have any similar validations. 
My question is does build CRD want that validation still? 

Thoughts or concerns @ImJasonH @mattmoor @pivotal-nader-ziada ?


pkg/apis/build/v1alpha1/build_defaults.go
```diff
@@ -0,0 +1,7 @@
+package v1alpha1
+
+// SetDefaults for build
+func (b *Build) SetDefaults() {
```
ImJasonH: Can you add `// TODO(shashwathi): Set default timeout to 10s` here?

pkg/apis/build/v1alpha1/build_template_validation.go
```diff
@@ -0,0 +1,125 @@
...
+		return err
+	}
+	if err := validateVolumes(b.Volumes); err != nil {
+		return err
+	}
+	if err := validateParameters(b.Parameters); err != nil {
+		return err
+	}
+	return nil
+	// return validatePlaceholders(tmpl.TemplateSpec().Steps)
```
ImJasonH: I think this was erroneously commented out.

pkg/apis/build/v1alpha1/build_validation.go
```diff
@@ -0,0 +1,182 @@
...
+
+// Validate for build spec
+func (bs *BuildSpec) Validate() *apis.FieldError {
+	if bs.Template == nil && len(bs.Steps) == 0 {
+		return apis.ErrMissingField("b.spec.template").Also(apis.ErrMissingField("b.spec.steps"))
+		//validationError("NoTemplateOrSteps", "build must specify either template or steps")
+	}
+	if bs.Template != nil && len(bs.Steps) > 0 {
+		return apis.ErrMissingField("b.spec.template").Also(apis.ErrMissingField("b.spec.steps"))
+		//return validationError("TemplateAndSteps", "build cannot specify both template and steps")
```
ImJasonH: These `return validationError` lines can just be deleted.

pkg/apis/build/v1alpha1/build_validation.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	if err := validateVolumes(append(bs.Volumes, volumes...)); err != nil {
+		return err
+	}
+
+	if err := validateSteps(bs.Steps); err != nil {
+		return err
+	}
+
+	// Need kubeclient to fetch template
```
ImJasonH: Yeah I think the current thinking is that we'll only do simple validation in the webhook, and leave it up to the controller to do full validation with resolved objects like the template, service account, secrets, etc.

I'm not sure there's a good story yet for surfacing validation errors outside of webhook checks though, we'd probably need to have some new condition type to express the `Invalid` state.

pkg/apis/build/v1alpha1/build_validation.go
```diff
@@ -0,0 +1,182 @@
...
+		BuildTemplateKind:
+		return nil
+	default:
+		return apis.ErrInvalidValue(string(b.Kind), apis.CurrentField)
+	}
+}
+
+// ValidateSecrets for build
+func (bs *BuildSpec) ValidateSecrets() *apis.FieldError {
+	// TODO: move this logic into set defaults
```
ImJasonH: +1 good call

shashwathi:
@ImJasonH 

I like the idea of 2 controllers for now. Updated code to split the validation between the 2 webhooks. Added tests and cleaned up bunch of common dependencies.

cmd/webhook/main.go
```diff
@@ -78,12 +83,27 @@ func main() {
...
+		WebhookName:    "webhook.build.knative.dev",
 	}
-	webhook.NewAdmissionController(kubeClient, buildClient, bldr, options, logger).Run(stopCh)
+
+	pkgcontroller := pkgwebhook.AdmissionController{
+		Client:  kubeClient,
+		Options: pkgoptions,
+		Handlers: map[schema.GroupVersionKind]pkgwebhook.GenericCRD{
+			v1alpha1.SchemeGroupVersion.WithKind("Build"):                &v1alpha1.Build{},
+			v1alpha1.SchemeGroupVersion.WithKind("ClusterBuildTemplate"): &v1alpha1.ClusterBuildTemplate{},
```
mattmoor: Can you confirm that the common infrastructure works properly with cluster-scoped resources, since AFAIK this is the first one we're trying :)

pkg/apis/build/v1alpha1/build_defaults.go
```diff
@@ -0,0 +1,20 @@
+package v1alpha1
+
+import (
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// DefaultTime is 10min
+const DefaultTime = 10 * time.Minute
```
mattmoor: nit: DefaultTimeout

cmd/webhook/main.go
shashwathi: Cool 👍 
I have added e2e test for build with template and build with cluster template as well. Going forward we will have better coverage for these features

pkg/apis/build/v1alpha1/build_defaults_test.go
```diff
@@ -0,0 +1,16 @@
+package v1alpha1
+
+import (
+	"testing"
+)
+
+func TestSetDefault(t *testing.T) {
```
ImJasonH: Can you also test that a build that specifies both values doesn't have its values overwritten?

pkg/apis/build/v1alpha1/build_template_defaults.go
```diff
@@ -0,0 +1,6 @@
+package v1alpha1
+
+// SetDefaults for build template
+func (b *BuildTemplate) SetDefaults() {
```
ImJasonH: Can we just put this into `build_template_types.go` ? This separate file doesn't add much.

pkg/apis/build/v1alpha1/build_template_validation_test.go
```diff
@@ -0,0 +1,127 @@
...
+					Image: "gcr.io/foo-bar/baz:latest",
+				}},
+			},
+		},
+	}, {
+		desc:   "long name",
+		reason: "long name",
+		tmpl: &BuildTemplate{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "*********************************************************************************************",
```
ImJasonH: `strings.Repeat("x", maxLength+1)` is self-documenting and self-updating if we update maxLength.

pkg/apis/build/v1alpha1/build_template_validation.go
```diff
@@ -0,0 +1,91 @@
...
+	name := meta.GetName()
+
+	if strings.Contains(name, ".") {
+		return &apis.FieldError{
+			Message: "Invalid resource name: special character . must not be present",
+			Paths:   []string{"name"},
+		}
+	}
+
+	if len(name) > 63 {
```
ImJasonH: Can we extract this to a const?

pkg/apis/build/v1alpha1/build_template_validation.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	return nil
+}
+
+func validateParameters(params []ParameterSpec) *apis.FieldError {
+	// Template must not duplicate parameter names.
+	seen := map[string]struct{}{}
+	for _, p := range params {
+		if _, ok := seen[p.Name]; ok {
+			return apis.ErrMultipleOneOf("ParamName")
```
ImJasonH: I don't think `MultipleOneOf` is the correct error, I think that error is intended to communicate "you can specify X or Y, but not both"

https://godoc.org/github.com/knative/pkg/apis#ErrMultipleOneOf

(`ErrMissingOneOf` is "you must specify X or Y, but not neither")

Maybe `ErrInvalidKeyName`? https://godoc.org/github.com/knative/pkg/apis#ErrInvalidKeyName

pkg/apis/build/v1alpha1/clutser_build_template_validation_test.go
```diff
@@ -0,0 +1,111 @@
+package v1alpha1
```
ImJasonH: typo in filename: `cluster_build_template_validation_test.go`

pkg/apis/build/v1alpha1/clutser_build_template_validation_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+func TestValidateClusterBuildTemplate(t *testing.T) {
+	hasDefault := "has-default"
+	for _, c := range []struct {
+		desc   string
+		tmpl   *ClusterBuildTemplate
+		reason string // if "", expect success.
+	}{{
+		desc: "Single named step",
+		tmpl: &ClusterBuildTemplate{
```
ImJasonH: I think we can condense both of these test tables into one, and avoid duplication:

```
func TestValidateBuildTemplate(t *testing.T) {
  for _, c := range []struct {
    desc string
    spec BuildTemplateSpec
    reason string
  } {{
   ...
  }} {
    t.Run("namespaced-"+c.desc, func(t *testing.T) {
      tmpl := &BuildTemplate{Spec: c.spec},
      tmpl.Validate()
      ...
    })
    t.Run("cluster-namespaced-"+c.desc, func(t *testing.T) {
      ctmpl := &ClusterBuildTemplate{Spec: c.spec},
      ctmpl.Validate()
      ...
    })
  }
}
```

pkg/apis/build/v1alpha1/build_template_validation_test.go
shashwathi: I missed my TODO there. Good catch

PR #380, opened 2018-09-28
Bump knative/pkg to master

vdemeester:
This is part kind of knative/pkg#97 as it updates knative/pkg to master and update the uses of `duck.VerifyType` (changed in knative/pkg#98)

cc @dprotaso @mattmoor

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -46,9 +46,8 @@ required = [
 
 [[constraint]]
   name = "github.com/knative/pkg"
-  # HEAD as of 2018-09-28
-  revision = "3155feb33006c8e0379a58237a819a36fae98c2d"
-
+  # HEAD as of 2018-10-02
+  revision = "8e471c3d74eda15fbb9405d076008e2cf2008548"
 
 [prune]
   go-tests = true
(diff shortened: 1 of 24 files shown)


Gopkg.toml
```diff
@@ -42,6 +42,11 @@ required = [
   name = "k8s.io/client-go"
   version = "kubernetes-1.11.0"
 
+[[override]]
```
shashwathi: There are 2 versions of pkg dependency in toml file. I think you can update the one under `constraint` to recent sha.

pkg/apis/build/v1alpha1/build_types_test.go
```diff
@@ -30,7 +30,9 @@ import (
 const bazelYAML = "testdata/cloudbuilders/bazel/cloudbuild.yaml"
 
 func TestBuildImplementsConditions(t *testing.T) {
-	duck.VerifyType(&Build{}, &duckv1alpha1.Conditions{})
+	if err := duck.VerifyType(&Build{}, &duckv1alpha1.Conditions{}); err != nil {
+		t.Error(err)
```
shashwathi: An error msg along with error would be helpful
eg: `t.Errorf("Expect Build to implement duck verify type: err %#v", err )`

Gopkg.toml
vdemeester: oh :sweat_smile: I didn't look below the `[[prune]]` options, my bad :sweat_smile:

PR #390, opened 2018-10-02
Move build-base to the build dir in GCR

adrcunha:
Let's put all knative/build images together, it's cleaner.

Part of https://github.com/knative/test-infra/issues/161

Bonus: separate building and pushing the base images, so you can safely use `--nopublish`.


adrcunha:
Oh, I have to push the base images to the new place for the integration tests to work... @ImJasonH the E2E tests shouldn't rely on these base images, right?

adrcunha:
Base image pushed.

/retest

PR #394, opened 2018-10-04
webhook: block execution

vdemeester:
Otherwise, the command will return immediately. Should fix the crashloop on the `build-webhook` pod on master

/cc @shashankmjain @bbrowning 


bbrowning:
Good catch!

/ok-to-test
/lgtm

PR #401, opened 2018-10-05
Bump apiVersion to apps/v1 for Deployments

jcrossley3:
This keeps build in sync with recent apiVersion changes in serving


jcrossley3:
Let's see...
/retest

PR #406, opened 2018-10-08
Fix TestBuildNotFoundFlow race condition

### LINKED ISSUE #404: Fix race condition in controller tests
Test [`TestBuildNotFoundFlow`](https://github.com/knative/build/blob/master/pkg/controller/build/controller_test.go) has race condition. 

sample failure logs: https://gubernator.knative.dev/build/knative-prow/pr-logs/pull/knative_build/395/pull-knative-build-unit-tests/1048242389451280385/

/area test-and-release

kind bug
/kind dev

## Expected Behavior
No race conditions in test

## Actual Behavior
Race conditions exist in controller test

  1. `go test -race ` multiple times

vdemeester:
Fixes #404 

## Proposed Changes

Add the reactor before starting the informers, making the "write"
operation on the fake client `ReactionChain` before any "read".

Let's see if it fixes the race on the CI… couldn't reproduce on my desktop…

### DIFF
--- a/pkg/controller/build/controller_test.go
+++ b/pkg/controller/build/controller_test.go
@@ -128,6 +128,15 @@ func TestBuildNotFoundFlow(t *testing.T) {
 
 	f.createServceAccount()
 
+	// induce failure when fetching build information in controller
+	reactor := func(action clientgotesting.Action) (handled bool, ret runtime.Object, err error) {
+		if action.GetVerb() == "get" && action.GetResource().Resource == "builds" {
+			return true, nil, fmt.Errorf("Inducing failure for %q action of %q", action.GetVerb(), action.GetResource().Resource)
+		}
+		return false, nil, nil
+	}
+	f.client.PrependReactor("*", "*", reactor)
+
 	stopCh := make(chan struct{})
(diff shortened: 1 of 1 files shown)


ImJasonH:
I'm also unable to reproduce this locally, even after hundreds of runs. I'd like to understand better how to reproduce this to make sure this is the fix, otherwise we're just going to keep making wild stabs until the race disappears (for now...).

This change looks fine insofar as I don't believe it will cause any _more_ problems, but I'm not sure how it would _fix_ any _existing_ races. But I wouldn't say it `Fixes #404` unless we can demonstrate that it does with a repro case.

vdemeester:
So my reasoning for this change is that the `fake` clients *kinda* have those race. There is no lock on the `ReactionChain` field on them, **and**, starting `k8sI` and `i` informers, may call `BuildList` before we added the reactor (or *while* we're adding the reactor)… It's probably triggered only in case of heavy load or such… But as we don't really need to start the informer before adding the reactor, it removes a potential race condition.

There might be some guidelines for those fakes (or an issues tracking these potential races)

PR #417, opened 2018-10-12
DEVELOPMENT: Apply variable expansion to volumes (#296)

### LINKED ISSUE #296: Parameter replacement for Volumes
## Expected Behavior
When template parameter is replaced with the value, only those under spec.steps are replaced. for example, volumeMounts will be replaced for its parameter value. 

However the ".spec.volumes" which is at the same level of ".spec.steps" is not replaced.

There is some use case when a template define its own volume, for example define a configMap volume to insert updated root certificate to the container.

JarreyZhou:
Fixes #296

## Proposed Changes

  * Apply variable expansion to volumes

**Release Note**

```release-note
```

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -223,11 +223,11 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:00f434f614520a4b48ef3f6d3ed5c7a1a4f89be1bed98c250711413f6da2eb51"
+  digest = "1:f42d23c4286880873a951329d5a6002fcb9ee9f7d6c5fd8263ea53df2391acee"
   name = "github.com/knative/test-infra"
   packages = ["."]
   pruneopts = "T"
-  revision = "c21d3a832727d5d4334e734c0878eac3a5a07ae7"
+  revision = "e7ddaaf7b3c285a538c6dfe061d59b73ace965cf"
 
 [[projects]]
   branch = "master"
--- a/test/e2e-tests.sh
+++ b/test/e2e-tests.sh
@@ -55,9 +55,7 @@ kubectl delete --ignore-not-found=true buildtemplates --all
 failed=0
 
 header "Running Go e2e tests"
-options=""
-(( EMIT_METRICS )) && options="-emitmetrics"
-report_go_test -tags e2e ./test/e2e/... -count=1 ${options} || failed=1
+go_test_e2e ./test/e2e/... || failed=1
 
 header "Running YAML e2e tests"
 if ! run_yaml_tests; then
--- a/test/presubmit-tests.sh
+++ b/test/presubmit-tests.sh
@@ -38,7 +38,9 @@ function unit_tests() {
 }
 
 function integration_tests() {
-  ./test/e2e-tests.sh
+  local options=""
+  (( EMIT_METRICS )) && options="--emit-metrics"
+  ./test/e2e-tests.sh ${options}
 }
 
 main $@
--- a/vendor/github.com/knative/test-infra/scripts/README.md
+++ b/vendor/github.com/knative/test-infra/scripts/README.md
@@ -1,3 +1,121 @@
 # Helper scripts
 
-This directory contains helper scripts used by Prow test jobs, as well and local development scripts.
+This directory contains helper scripts used by Prow test jobs, as well and
+local development scripts.
+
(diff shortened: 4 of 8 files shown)


pkg/builder/common.go
```diff
@@ -100,7 +100,26 @@ func ApplyReplacements(build *v1alpha1.Build, replacements map[string]string) *v
...
+	// Apply variable expansion to volumes.
+	if build.Spec.Volumes != nil {
+		volumeBytes, err := json.Marshal(build.Spec.Volumes)
+		if err != nil {
+			return nil, err
+		}
+
+		volumes := []corev1.Volume{}
+
+		err = json.Unmarshal([]byte(applyReplacements(string(volumeBytes))), &volumes)
```
ImJasonH: This isn't a very elegant way to apply these replacements, and it could conceivably result bugs with unexpected replacements.

I think it would be much better to iterate through the build's volumes and `applyReplacements` to each volume's name. Do we expect to need to apply replacements inside volume sources too?

JarreyZhou: you are right, it will cause unexpected replacement. By just replacing volume's name makes no sense. In #296 , i think that he wants to replace something in volume sources like configMap's name.

ImJasonH:
In that case maybe we need to step back and clarify what we're trying to support.

If it's mainly to support replacements in config map volumes at this time, we can add support for _only_ those volume types for now, and document the limitation. If there are other easy/worthwhile volume types to support, we can do those too.

I just want to avoid doing all the work to support and test aaaallll the types of volumes, since most will never be used.

Does that make sense?

JarreyZhou:
Should I just replace volume's name this time, and then add replacement to volumeSource if needed.

ImJasonH:
If applying replacements to the volume name is sufficient to close #296 then sure. But it sounded like  you had a specific use case in mind that wanted replacements for config map values, etc.

If that use case is more of a hypothetical one, then maybe we can wait until the use case arises in real usage, and address it based on the extent of the real user need. In that case, we'd keep the issue open and just remove it from the 0.2 milestone.

JarreyZhou:
I think I may need more communication with the issuer to ensue.

JarreyZhou:
@ImJasonH ， what the issuer want is to replace configMap's name in volumes by just providing params in build, so this user case need to replace something in volumeSource.

JarreyZhou:
can I implement like this (using reflect and replace all the string values in volumes) 
```
func Replace(v reflect.Value, replace func(string) string) {

	switch v.Type().Kind() {
	case reflect.String:
		v.SetString(replace(v.String()))
	case reflect.Interface:
		Replace(v.Elem(), replace)
	case reflect.Struct:
		ReplaceStruct(v, replace)
	case reflect.Map:
		ReplaceMap(v, replace)
	case reflect.Slice:
		ReplaceSlice(v, replace)
	case reflect.Array:
		ReplaceSlice(v, replace)
	case reflect.Ptr:
		Replace(v.Elem(), replace)
	default:
	}
	return
}

func ReplaceStruct(v reflect.Value, replace func(string) string) {

	for i := 0; i < v.NumField(); i++ {
		Replace(v.Field(i), replace)
	}
}

func ReplaceMap(v reflect.Value, replace func(string) string) {
	if v.IsNil() {
		return
	}

	keys := v.MapKeys()
	for _, key := range keys {
		Replace(v.MapIndex(key), replace)
	}
}

func ReplaceSlice(v reflect.Value, replace func(string) string) {
	if v.IsNil() {
		return
	}

	for i := 0; i < v.Len(); i++ {
		Replace(v.Index(i), replace)
	}
}

```

ImJasonH:
@JarreyZhou I'm a bit worried that this will have unexpected consequences when applied to any arbitrary VolumeSource, including any future VolumeSource implementation where `${foo}` might be a valid string in that context and shouldn't be replaced by us.

Instead, maybe we should start with only replacing in a small set of volume types we know we want to support (config maps, for instance), and document the types of volumes where replacements will be applied. If users need new types of volumes to be supported, we can add them at that time based on their use case. If it starts to look like users want to be able to replace across all types of volumes, we can do the fully general reflective solution you've outlined above.

Does that sound reasonable to you?

JarreyZhou:
ok,  it make sense. I will add recode it and add some docs to  https://github.com/knative/docs/tree/master/build

PR #418, opened 2018-10-12
Add tests for credentials package

shashwathi:
## Proposed Changes

  * Add more tests for credentials package


shashwathi:
I am unable to reproduce the error locally. I will trigger again to see if this is flake.

/test pull-knative-build-unit-tests

shashwathi:
Order of flags is not guaranteed and that causes test to be flaky. I will fix that and remove hold

shashwathi:
@ImJasonH : Let me follow up with a PR to address that comment. I agree it removes uncertainty.

PR #419, opened 2018-10-12
Sort docker and git credentials

shashwathi:
## Proposed Changes

  * If build uses secret that has multiple annotations then they are sorted alphabetically
  * Reduces random picking of key.

### DIFF
--- a/pkg/credentials/initialize.go
+++ b/pkg/credentials/initialize.go
@@ -18,6 +18,8 @@ package credentials
 
 import (
 	"fmt"
+	"sort"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 )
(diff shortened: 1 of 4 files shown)


pkg/credentials/initialize.go
```diff
@@ -41,3 +43,17 @@ type Builder interface {
 func VolumeName(secretName string) string {
 	return fmt.Sprintf("%s/%s", VolumePath, secretName)
 }
+
+// SortAnnotations alphabetically
```
ImJasonH: Incomplete comment here

shashwathi:
/assign @ImJasonH 

Let me know if I addressed your concern correctly.

pkg/credentials/initialize.go
```diff
@@ -41,3 +43,17 @@ type Builder interface {
...
+func SortAnnotations(secrets map[string]string, annotationPrefix string) []string {
+	var mk []string
+	i := 0
+	for k, v := range secrets {
+		if strings.HasPrefix(k, annotationPrefix) {
+			mk = append(mk, v)
+			i++
```
ImJasonH: Looks like you don't need `i` for anything here

shashwathi:
@ImJasonH  : Addressed your comments

PR #415, opened 2018-10-11
Extract the ApplyReplacements logic from ApplyTemplate.

dlorenc:
This functionality will be used in build-pipeline as well, and reusing
this logic will help ensure consistency.

Should be used for https://github.com/knative/build-pipeline/issues/64

## Proposed Changes

  * Extract logic from ApplyTemplate into a new, public function `ApplyReplacements`

**Release Note**
```release-note
NONE
```

/lint

PR #420, opened 2018-10-15
[WIP] Migrate to reconciler for builds

### LINKED ISSUE #393: Use as much of knative/pkg as is feasible:
- webhook validation
- controller -> reconciler

ImJasonH:
Fixes #393

## Proposed Changes

  * Use pkg reconciler pattern to reconcile builds
  * Delete `pkg/builder/...` -- only the on-cluster implementation remains, it converts Builds into Pods and creates them -- the `nop` builder impl is no longer necessary
  * Validate build after applying template

## WIP

This PR doesn't currently:
  * take into account build timeouts
  * export `ApplyReplacements` like #415 needs

Looking for preliminary feedback on this approach before adding timeout handling.

**Release Note**

```release-note
Use knative/pkg reconciler pattern for builds
```


pkg/reconciler/build/build.go
```diff
@@ -44,14 +53,16 @@ type Reconciler struct {
...
+	buildTemplatesLister        listers.BuildTemplateLister
+	clusterBuildTemplatesLister listers.ClusterBuildTemplateLister
 
 	// Sugared logger is easier to use but is not as performant as the
 	// raw logger. In performance critical paths, call logger.Desugar()
 	// and use the returned raw logger instead. In addition to the
 	// performance benefits, raw logger also preserves type-safety at
 	// the expense of slightly greater verbosity.
-	Logger *zap.SugaredLogger
+	logger *zap.SugaredLogger
```
shashwathi: You could consider including Base Reconciler inside build reconciler like [this](https://github.com/knative/serving/blob/master/pkg/reconciler/v1alpha1/service/service.go#L44-L51). You get declaration cruft for free

pkg/reconciler/build/build.go
```diff
@@ -107,29 +122,205 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
+	}
+	return err
+}
+
+func (c *Reconciler) reconcile(ctx context.Context, b *v1alpha1.Build) error {
+	logger := logging.FromContext(ctx)
+
+	// If the build doesn't define a pod name, it means it's new. We should
+	// apply any template and create its pod.
+	if b.Status.Cluster == nil || b.Status.Cluster.PodName == "" {
```
shashwathi: How about this

```
func (c *Reconciler) reconcile(ctx context.Context, b *v1alpha1.Build) error {
pod, err := getPod(b)
if err.IsNotFound(err) {
createPod() // create pod and apply template
} else if err != nil  {
  return err
}
// sync status
```

shashwathi: First thing to check would be build status whether its completed or not. If build is completed, ` b.Status.Cluster.PodName` will be set but `GetPod()` will fail with not found error.

pkg/reconciler/build/build.go
mattmoor: Unfortunately we define that base in `knative/serving` :-/

PR #427, opened 2018-10-17
DEVELOPMENT: Apply variable expansion to volumes (knative#296)

JarreyZhou:
Fixes #
    #296
## Proposed Changes

  * Apply variable expansion to volumes

**Release Note**

```release-note
```


pkg/builder/common_test.go
```diff
@@ -720,6 +720,49 @@ func TestApplyReplacements(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "replacement in volumes",
+			args: args{
+				build: &v1alpha1.Build{
+					Spec: v1alpha1.BuildSpec{
+						Volumes: []corev1.Volume{
+							{
```
ImJasonH: nit: You can save two lines of vertical space and a tab-width of horizontal space by collapsing some braces:

```
Volumes: []corev1.Volume{{
  Name: "${name}",
  VolumeSource: corev1.VolumeSource{
    ...
  },
}},

File: pkg/builder/common.go [hunk 2]
```diff
@@ -100,6 +100,15 @@ func ApplyReplacements(build *v1alpha1.Build, replacements map[string]string) *v
 			steps[i].Env = applyEnvOverride(steps[i].Env, buildTmpl.Env)
 		}
 	}
+
+	// Apply variable expansion to volumes fields.
+	if volumes := build.Spec.Volumes; volumes != nil && len(volumes) > 0 {
+
```
ImJasonH: nit: Unnecessary blank line here

JarreyZhou: fixed

JarreyZhou:
cc @ImJasonH @shashwathi

shashwathi:
@JarreyZhou @ImJasonH 

I think there is value in adding go e2e test which refers to volume replacement of configmap that does not exist? Also verify that build fails for expected reason.

WDYT?

JarreyZhou:
@shashwathi 
I will do that. But I do not no how to compare the failed reason. Write it in labels ?? I just succeeded and filed can be set.

shashwathi:
@JarreyZhou :
If build refers to a non existent volume I would expect the pod failure to be bubbled up in build status msg. Its a bit of hack but you could verify the build status msg to contain something like `Volume A is not present`. If the last part of status verification is too much work then feel free to just test that build fails for now. Its still a good test I think.

JarreyZhou:
@shashwathi , can I add this failed test in the same folder with the succeeded one?

JarreyZhou:
@shashwathi , can I add this failed test in the same folder with the succeeded one?

JarreyZhou:
@shashwathi , can I add this failed test in the same folder with the succeeded one?

JarreyZhou:
@shashwathi , can I add this failed test in the same folder with the succeeded one? 

> @JarreyZhou :
> If build refers to a non existent volume I would expect the pod failure to be bubbled up in build status msg. Its a bit of hack but you could verify the build status msg to contain something like `Volume A is not present`. If the last part of status verification is too much work then feel free to just test that build fails for now. Its still a good test I think.

JarreyZhou:
@shashwathi , can I add this failed test in the same folder with the succeeded one?

shashwathi:
@JarreyZhou : There is [e2e test file](https://github.com/knative/build/blob/master/test/e2e/invalid_test.go) which consist of invalid builds. You can consider adding it to that list.

JarreyZhou:
@shashwathi , sorry for check this message so late. I think this should be a failed test not a invalid test. So i will add the failed test case in simple_test.go

JarreyZhou:
@shashwathi , should I check if the configmap is exist before create pod , as pod with configmap not exist will be on Init status all the time.

shashwathi:
@JarreyZhou : Thats a good check. If test result is affected by existence of configmap then lets validate configmap doesn't exist before creating the build.

shashwathi:
@JarreyZhou 

`clients` in e2e test contains `kubeClient`. You could consider doing 

```
if err := clients.kubeclient.corev1().ConfigMap(namespace).Delete(test_configmap_name) ;  !kuberrors.IsNotFound(err) {
       logger.Errorf("Failed to delete test configMap: %v", err)
}
```

WDYT?

PR #430, opened 2018-10-18
Fix e2e:TestInvalidBuild flakiness 

dprotaso:
My current theory is the webhook is not registered/running when the tests start.

This waits until all pods in the knative-build namespace have their 'Ready' condition set to true.

### DIFF
--- a/test/e2e-tests.sh
+++ b/test/e2e-tests.sh
@@ -29,6 +29,16 @@
 source $(dirname $0)/../vendor/github.com/knative/test-infra/scripts/e2e-tests.sh
 source $(dirname $0)/e2e-common.sh
 
+function wait_for_running_build_pods() {
+  echo "Waiting for knative build pods to be ready..."
+  local pods=$(kubectl get pods -o name -n knative-build)
+
+  for pod in $pods
+  do
+    kubectl wait --for=condition=Ready -n knative-build "$pod"
+  done
+  echo "Knative build pods are ready"
+}
 
 # Script entry point.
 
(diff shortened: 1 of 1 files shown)


dprotaso:
/hold

I want to run the integration test numerous times

dprotaso:
@ImJasonH I don't think serving has these types of tests

test/e2e-tests.sh
```diff
@@ -50,6 +60,8 @@ set +o pipefail
 kubectl delete --ignore-not-found=true builds --all
 kubectl delete --ignore-not-found=true buildtemplates --all
 
+wait_for_running_build_pods
```
adrcunha: I think that `wait_until_pods_running knative-build` would achieve the same thing with less code.

Interesting, looks like the script never really checked for the pods being up, like serving or eventing do (or maybe it was lost in a refactoring?).

dprotaso: Oh thanks for pointing that out - i'll make the changes

adrcunha:
By the way, if you want to test it several times just like Prow:

```
export PROJECT_ID=dprotaso_awesome_gcp_project; for i in {1..50}; do ./test/e2e-tests.sh ; done
```

dprotaso:
/hold cancel

seems ok

dprotaso:
@adrcunha @ImJasonH mind giving this another 
/lgtm

test/e2e-tests.sh
adrcunha: Funny, I don't see the changes here. Was it intentional?

shashwathi:
@adrcunha : My bad. It looked from comments as though everything was sorted.

adrcunha:
No worries, Dave fixed it in https://github.com/knative/build/pull/431

PR #433, opened 2018-10-19
include missing git packages in build-base

dprotaso:
This addresses failing e2e tests where the git source uses https


dprotaso:
Closing in favour of the quick revert to fix e2e tests not being blocked.

PR #434, opened 2018-10-19
Don't cleanup bazel on Prow for releases

adrcunha:
Otherwise it will erase the test results.


dprotaso:
@adrcunha with this revert https://github.com/knative/build/commit/17e8cf8417e1ef3d29bd465d4f45ad19dd3a3f2c you don't need your changes anymore

PR #446, opened 2018-10-26
Migrate to reconciler pattern for Build 

### LINKED ISSUE #393: Use as much of knative/pkg as is feasible:

ImJasonH:
Fixes #393

## Proposed Changes

  * Copy code out of `pkg/controller/build/` and into `pkg/reconciler/build/`
  * Make tests pass
  * That's pretty much it.

This is a smaller, more mechanical change than #420 which attempted to roll in a refactor which would have let us drop `builder.Interface`. I'd still like to get rid of that abstraction and just always interact with a `kubeClient` (which might be fake in tests), but I'll get to that some time after 0.2.

/assign shashwathi
/assign mattmoor

**Release Note**
```release-note
Migrate to reconciler pattern for Build 
```


shashwathi:
I definitely like the scope of this PR. Great work @ImJasonH

pkg/reconciler/build/build.go
```diff
@@ -109,29 +125,151 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
+	build = build.DeepCopy()
+
+	// Validate build
+	if err = c.validateBuild(build); err != nil {
+		c.Logger.Errorf("Failed to validate build: %v", err)
+		return err
+	}
+
+	// If the build's done, then ignore it.
+	if !builder.IsDone(&build.Status) {
```
mattmoor: feels worth inverting this:
```go
if builder.IsDone(&build.Status) {
  return nil
}
```

pkg/reconciler/build/build.go
```diff
@@ -109,29 +125,151 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
+	}
+	build.Status = *status
+	if _, err := c.updateStatus(build); err != nil {
+		c.Logger.Errorf("Error updating build status: %v", err)
+		return err
+	}
 	return nil
 }
+
+func (c *Reconciler) updateStatus(u *v1alpha1.Build) (*v1alpha1.Build, error) {
```
mattmoor: At some point, you should pick up the pattern from `knative/serving` where we have a `Reconcile()` and `reconcile()` and the former just calls the latter, and then calls `updateStatus` regardless of whether the latter returned an `error`.

pkg/reconciler/build/build.go
```diff
@@ -109,29 +125,151 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
+				return err
+			}
+			// TODO: Validate build except steps+template
+			b, err := c.builder.BuildFromSpec(build)
+			if err != nil {
+				return err
+			}
+			op, err := b.Execute()
+			if err != nil {
+				build.Status.SetCondition(&duckv1alpha1.Condition{
```
mattmoor: Another thing to pick up is the use of higher-level functions (I actually removed `setCondition` from our types).

pkg/reconciler/build/build.go
ImJasonH: Good call. I can also remove another layer of indentation!

pkg/reconciler/build/build.go
ImJasonH: Yeah, I like that pattern. I think that pattern opens up some more cleanup opportunity that I think I'd like to keep in a separate PR.

pkg/reconciler/build/build.go
ImJasonH: Yeah, that's definitely a good idea. Probably easier to do once we remove the builder split, since some `SetCondition` calls are in there. It would be much easier to make this change if everything was linear.

My goal with this PR was to change as little code as possible, even though there are plenty of threads to pull. But getting this in before 0.2 is the priority I think. More cleanups next week! 🎉

Issue #423, opened 2018-10-16
Support for corporate environments - CA certs, HTTP proxies


DanielJonesEB:
We're attempting to use Riff (which in turn uses Knative Build) inside a corporate network.

We need to be able to provide:

* a CA certificate to steps in the build process, such that processes in those build steps can make HTTPS communications to internal infrastructure (GitLab, BitBucket, etcetera) serving TLS certificates signed by that CA.
* environment variables (`HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`, `http_proxy`, `https_proxy`, `no_proxy`) so should be exported in build containers such that processes can communicate via HTTP proxies where appropriate (typically external services like GitHub), and _not_ via HTTP proxies for specified IPs/hostnames (for internal services).

We're using Riff which presumably uses Knative Build to create a pod spec and schedule it. The initContainers in this pod need to have access to the env vars and the CA cert.

The CA cert _may_ require additional steps. We've been creating a manual test case in a pod of our own creation, whereby we add the base64-encoded CA cert as a Secret in Kubernetes, then mount that into our container. We're probably _doing it wrong_ but in an Ubuntu-based image, mounting the cert directly into `/etc/ssl/certs/` doesn't work and instead we need to mount it into `/usr/local/share/ca-certificates/` and then run the `update-ca-certificates` command. I'm not sure if this process is universal across Linux distributions; if not then I can understand why supporting this might be impractical.

I'm not super-familiar with Knative Build so I'm not sure what the UX should look like. 

cc @andrewdriver123

## Steps to Reproduce the Problem

  1.
  2.
  3.

shashwathi:
@DanielJonesEB : 

- Setting environment variables on build step here is an [example](https://github.com/knative/build/blob/master/test/custom-env/test.yaml) . You could also consider using configMap with all `HTTP`/`HTTPS` variables and then pass the configMap references to build steps. Here is the config map [example](https://github.com/knative/build/tree/master/test/configmap)
- [Example](https://github.com/knative/build/tree/master/test/secret-volume) to mount CA certificates volume in build.
- And for your last question you could consider adding a step in build to mount the certificate volume and run the command `update-ca-certificates` before using the same volume in next step(in theory this should work).

ImJasonH:
I think the bigger issue might be how to set those configurations for the step we prepend that fetches the specified source. If `HTTP_PROXY` is needed to fetch from the Git repo, for instance, there's currently no way to inject that.

[...]

PR #452, opened 2018-11-01
Add support for setting environment variables on git init container

shashwathi:
This change will allow build to use git repositories behind corporate firewalls
or any proxy sites.

Partially addresses #423 

## Proposed Changes

  * Git source takes `env` key. 
  * All variables under `env` will be set on init container
 

**Release Note**

```
Set environment variables like proxy variables on git init containers.
```

* No e2e test in this PR
  
   I have added only unit tests for this feature. I am not aware of any git URLs which is behind firewall so if anybody has ideas about adding e2e test which uses environment variable feel free to comment on this PR. I am happy to update with integration test.

cc @ImJasonH @mattmoor


shashwathi:
@DanielJonesEB @andrewdriver123 let me know if this solution helps you with #423 issue. I would love to hear your thoughts.

mattmoor:
@shashwathi with this can the url contain env vars?  if so, you can write an e2e test :)

shashwathi:
@mattmoor 
I like that idea. Gives me an opportunity to test for sure. Outside of testing, Is there any real use case for templating url parts from env vars?
ex: Is there a world where github URL would be fetched from secret/configMap?

fejta:
Wouldn't it be better for source to use a higher level abstraction than env var?

I would think this provides too tight a coupling with the internal of the git container. It would seem better for people who need this to use a custom clone container, which already allows setting env vars.

PR #453, opened 2018-11-05
Add optional tags for build spec

shashwathi:
This is recommended api convention by k8s.

Ref: https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#optional-vs-required

## Proposed Changes

  * Add `+optional` tags

### DIFF
--- a/pkg/apis/build/v1alpha1/build_types.go
+++ b/pkg/apis/build/v1alpha1/build_types.go
@@ -222,11 +222,11 @@ type BuildStatus struct {
 
 	// StartTime is the time the build is actually started.
 	// +optional
-	StartTime metav1.Time `json:"startTime,omitEmpty"`
+	StartTime *metav1.Time `json:"startTime,omitEmpty"`
 
 	// CompletionTime is the time the build completed.
 	// +optional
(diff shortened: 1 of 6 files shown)


pkg/apis/build/v1alpha1/build_types.go
```diff
@@ -198,25 +209,34 @@ const (
...
+	// +optional
 	Cluster *ClusterSpec `json:"cluster,omitempty"`
+
 	// Google provides additional information if the builder is Google.
+	// +optional
 	Google *GoogleSpec `json:"google,omitempty"`
 
 	// StartTime is the time the build is actually started.
+	// +optional
 	StartTime metav1.Time `json:"startTime,omitEmpty"`
```
pivotal-nader-ziada: I think `StartTime` and `CompletionTime ` should be a pointer

see an example here:

https://github.com/kubernetes/kubernetes/blob/7f23a743e8c23ac6489340bbb34fa6f1d392db9d/staging/src/k8s.io/apiserver/pkg/apis/example/types.go#L70

shashwathi: I have updated PR to change startTime and completeTime to be pointer

PR #457, opened 2018-11-08
Move pkg/builder/cluster/convert to pkg/reconciler/build/resources

ImJasonH:
This moves us a step closer to removing `builder.Interface`, and makes us a bit more consistent with Knative best practices.

As a next step, I'll remove the `builder.Interface` and simply always create pods. In tests, these pods will be created against a fake k8s client.

## Proposed Changes

  * Essentially, `mv pkg/builder/cluster/convert.go pkg/reconciler/build/resources/pod.go`, etc.
  * Remove `pkg/builder/validation` package and use `github.com/knative/pkg/apis.FieldError` instead, again to become more consistent.
  * Remove `ToCRD` which was never used outside of testing.

**Release Note**
```release-note
NONE
```


shashwathi:
> Remove ToCRD which was never used outside of testing.

Awesome I like that its not in source code anymore. If this function has value in testing does it need to move into testing helpers instead of removing it completely?

Better way to frame the question "what is the value `TestRoundtrip` provides?" 
- It checks SA and init container volumeMounts. I do not know whether so many testdata yaml files provides any testing value since we do not inspect the pod spec. 
- If we choose to move away from `ToCRD`then probably then retain yaml files which makes sense.
I guess I am trying to say if we are doing effort to clean up `ToCRD` function then related files could be cleaned up too.

ImJasonH:
That's an excellent question. I think I've become convinced that `TestRoundtrip` doesn't add much, at least not enough to warrant all those testdata files. Without doing the full roundtrip (which doesn't test anything useful), it essentialy just tests that those YAML files are parseable and can make it through `MakePod` without error.

Instead, I've added the SA and secret volume mount stuff to `TestMakePod` and deleted `TestRoundtrip` and all the testdata files. Even more red! :tada:

PR #464, opened 2018-11-09
Remove builder interface, linearize pod creation and watching

ImJasonH:
## Proposed Changes

  * Remove the `builder.Interface` which effectively only had a single implementation for on-cluster builds
  * Linearize pod creation, so it can be read and understood more clearly from top-to-bottom
  * Remove `Operation`, callbacks, mutexes, etc.

With this change, we'll listen for all pod events on the cluster, and when we see one for a pod owned by a Build, we'll update that build with the Pod's status.

**Release Note**
```release-note
NONE
```


ImJasonH:
The failing test is the YAML test for `test-with-low-timeout`:
```
I1109 22:02:23.123] test-with-low-timeout     Succeeded   Unknown   2018-11-09T21:56:46Z   <none>     <none>
```

It looks like the timeout isn't kicking in and the build continues executing ~forever. Oddly this doesn't seem to happen when I run e2e-tests.sh locally. I'll investigate more and unhold this.

/hold

pkg/reconciler/build/resources/pod.go
```diff
@@ -329,8 +329,7 @@ func MakePod(build *v1alpha1.Build, kubeclient kubernetes.Interface) (*corev1.Po
 			// We execute the build's pod in the same namespace as where the build was
 			// created so that it can access colocated resources.
 			Namespace: build.Namespace,
-			// Ensure our Pod gets a unique name.
-			GenerateName: fmt.Sprintf("%s-", build.Name),
+			Name:      fmt.Sprintf("pod-for-%s", build.Name),
```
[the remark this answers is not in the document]
ImJasonH: Good call, done.

FWIW, the reason I dropped `GenerateName` is because the fake k8s implementation doesn't generate names, and so pods were always created as `""`. 😠

cmd/controller/main.go
```diff
@@ -104,11 +103,9 @@ func main() {
 	clusterBuildTemplateInformer := buildInformerFactory.Build().V1alpha1().ClusterBuildTemplates()
 	imageInformer := cachingInformerFactory.Caching().V1alpha1().Images()
 
-	bldr := onclusterbuilder.NewBuilder(kubeClient, kubeInformerFactory, logger)
-
 	// Build all of our controllers, with the clients constructed above.
-	controllers := []controller.Interface{
-		build.NewController(logger, kubeClient, buildClient, buildInformer, buildTemplateInformer, clusterBuildTemplateInformer, bldr),
+	controllers := []*controller.Impl{
+		build.NewController(logger, kubeClient, kubeInformerFactory, buildClient, buildInformer, buildTemplateInformer, clusterBuildTemplateInformer),
```
mattmoor: The pattern we're using constructs the Pod informer here and passes it in, so that it can be included in the synchronization below.

pkg/reconciler/build/build.go
```diff
@@ -107,8 +105,13 @@ func NewController(
...
 	})
 
-	// TODO(mattmoor): Set up a Pod informer, so that Pod updates
-	// trigger Build reconciliations.
+	// Set up a Pod informer, so that Pod updates trigger Build
+	// reconciliations.
+	kubeinformers.Core().V1().Pods().Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    r.addPodEvent,
+		UpdateFunc: r.updatePodEvent,
+		DeleteFunc: r.deletePodEvent,
```
mattmoor: For this to be properly level-based, these should use `impl.EnqueueControllerOf` (possibly wrapped with `controller.PassNew`).  For a Pod informer, you also certainly want to use a Filter on the OwnerReference Kind being Build.

In level-based reconciliation, all of the logic is oriented in terms of reconciling Build keys (`Reconcile()`), and these triggers are very thin, they just figure out what Build to queue and enqueue it.

ImJasonH: Thanks for that tip. I tried switching to this and had some trouble in tests getting a signal that the enqueued update happened. Do you happen to have a link to how you handle this in Serving?

Also, good idea filtering for build-owned pods. How do I express that filter to `podsLister`? [`NewSharedInformerFactoryWithOptions`](https://godoc.org/k8s.io/client-go/informers#NewSharedInformerFactoryWithOptions) with a [`TweakListOptions`](https://godoc.org/k8s.io/client-go/informers#WithTweakListOptions) to specify a `FieldSelector`?

ImJasonH:
/hold

I have quite a few unpushed changes, I'm thinking this might be better as more smaller PRs. Unfortunately life stuff is taking up a lot of time lately, I probably can't promise much progress until after the Thanksgiving holiday.

ImJasonH:
/hold cancel

This change is ready now, passing unit and integration tests. Unfortunately it's pretty big, sorry about that, my intention had been to split it up but it looked like it was all pretty tightly intertwined. The good news is I think this is a _lot_ easier to read and understand (and test!) after this change. It should now be pretty straightforward to forklift this into the `TaskRun` controller and make it responsible for creating Pods directly.

pkg/reconciler/build/build.go
```diff
@@ -137,63 +148,76 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
+
+	return c.updateStatus(build)
+}
+
+func (c *Reconciler) updateStatus(u *v1alpha1.Build) error {
+	newb, err := c.buildclientset.BuildV1alpha1().Builds(u.Namespace).Get(u.Name, metav1.GetOptions{})
+	if err != nil {
+		return err
 	}
+	newb.Status = u.Status
```
shashwathi: We could consider checking if there is any change in status object before updating it.
This could be changed in further PR too.

test/e2e/e2e.go
```diff
@@ -164,19 +174,20 @@ func (c *buildClient) watchBuild(name string) (*v1alpha1.Build, error) {
...
-					return b, errors.New("build failed")
+					return b, errBuildFailed
 				case corev1.ConditionUnknown:
 					continue
 				}
 			}
 		}
 	}
-	return nil, errors.New("watch ended before build completion")
+	return latest, errWatchTimeout
```
shashwathi: I like this classification of errors.

vdemeester:
oh @ImJasonH needs a rebase :pray:

PR #466, opened 2018-11-09
Add bash helper function to get pod logs

shashwathi:
Hope this helps with debugging test failures. 

Credit to @markusthoemmes work on knative/serving [PR]( https://github.com/knative/serving/pull/2432)

### DIFF
--- a/test/e2e-tests.sh
+++ b/test/e2e-tests.sh
@@ -29,6 +29,25 @@
 source $(dirname $0)/../vendor/github.com/knative/test-infra/scripts/e2e-tests.sh
 source $(dirname $0)/e2e-common.sh
 
+# Helper functions.
+function dump_app_logs() {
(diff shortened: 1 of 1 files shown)


test/e2e-tests.sh
```diff
@@ -29,6 +29,25 @@
 source $(dirname $0)/../vendor/github.com/knative/test-infra/scripts/e2e-tests.sh
 source $(dirname $0)/e2e-common.sh
 
+# Helper functions.
+function dump_app_logs() {
```
adrcunha: This is nice. We should move this to the helper script so every repo can use it.

shashwathi: I agree completely

adrcunha: Filed https://github.com/knative/test-infra/issues/252 so we don't forget. :)

PR #467, opened 2018-11-10
Delete unused testdata directory, useless test, unused test methods

ImJasonH:
## Proposed Changes

  * Remove `TestParsing` from `build_types_test.go` which for-some-reason tested that a GCB build config could be (partially...?) parsed into a Build CRD. It only tested it with one of the testdata YAMLs anyway, and nobody cares about that feature.
  * Remove `pkg/apis/build/v1alpha1/testdata/**` which was unused except for that test ☝️ 
  * Remove `pkg/buildtest/{data,yaml}.go` which were unused except for that test ☝️ 
  * Remove `pkg/buildtest/json.go` which was just entirely unused.

After this change, the only thing in `pkg/buildtest` is `wait.go` which is only used by `pkg/builder/cluster/builder_test.go`, which will also be deleted in #464 

**Release Note**
```release-note
NONE
```


pkg/apis/build/v1alpha1/build_types_test.go
```diff
@@ -113,3 +92,46 @@ func TestBuildGroupVersionKind(t *testing.T) {
 		t.Errorf("GetGroupVersionKind mismatch; expected: %v got: %v", expectedKind, b.GetGroupVersionKind().Kind)
 	}
 }
+
+// dataAs interprets the YAML contents of the file at the given path as the
+// given type.
+func dataAs(relpath string, obj interface{}) error {
```
shashwathi: Where is this function being used?

ImJasonH: Oops, nothing uses it anymore. Originally this PR was to move `buildtest.DataAs` into this test package, but then I realized we didn't need it at all, so we can just get rid of all of it! 🗑

PR #472, opened 2018-11-20
Increase timeout for TestSimpleBuild

mgencur:
This test randomly fails on our OpenShift instance due to timeout. I think increasing the timeout to 120 is sufficient in general (it passes constantly in our case).

### DIFF
--- a/test/e2e/simple_test.go
+++ b/test/e2e/simple_test.go
@@ -83,7 +83,7 @@ func TestSimpleBuild(t *testing.T) {
 			Name:      buildName,
 		},
 		Spec: v1alpha1.BuildSpec{
-			Timeout: &metav1.Duration{Duration: 40 * time.Second},
+			Timeout: &metav1.Duration{Duration: 120 * time.Second},
 			Steps: []corev1.Container{{
 				Image: "busybox",
 				Args:  []string{"echo", "simple"},


ImJasonH:
@mgencur Thanks for the PR. Do you have any idea what's taking so long on your cluster? Our Prow tests run this test every day and the build never takes more than 8s ([link](https://testgrid-dot-knative-tests.appspot.com/knative-build#latency)).

If the added latency is coming from something we can improve (such as by producing smaller builder images, or pre-caching them), I think we should definitely improve those. There's basically no reason this test should take more than 40s.

mgencur:
@ImJasonH I'm not sure but I think it's caused by limited resources and the fact that not all Builds (Pods) can run at the same time. I was also thinking about making this configurable. But it's used only in one place so I thought just increasing the default timeout could fix it without extra effort.

test/e2e/simple_test.go
```diff
@@ -35,6 +35,8 @@ import (
 	"github.com/knative/build/pkg/apis/build/v1alpha1"
 )
 
+const defaultBuildTimeout = 120 * time.Second
```
ImJasonH: I think it might be better to keep this inlined in the test, unless you plan to update all the tests to specify this timeout value. Otherwise, "default" might be a bit confusing, since it's only actually used in one place.

mgencur: OK. Thanks! The timeout is now inlined.

PR #479, opened 2018-11-27
Publish nightly knative/build releases to knative-nightly

adrcunha:
Part of https://github.com/knative/serving/issues/2555.

Bonus: updates `test-infra` to include the latest fixes and features, while also updating the Gopkg files so the cleanup workaround in `update-deps.sh` is not necessary anymore.


.ko.yaml
```diff
@@ -1,3 +1,3 @@
 baseImageOverrides:
-  github.com/knative/build/cmd/creds-init: gcr.io/knative-releases/github.com/knative/build/build-base:latest
-  github.com/knative/build/cmd/git-init: gcr.io/knative-releases/github.com/knative/build/build-base:latest
+  github.com/knative/build/cmd/creds-init: gcr.io/knative-nightly/github.com/knative/build/build-base:latest
```
ImJasonH: Won't this make real releases base these images on the nightly base images? Is that a problem?

adrcunha: Yes, it would, but that's how it is now. Nightlies are supposed to be stable because we run tests before publishing. We can base these images on the latest versioned release if thats better.

vdemeester: oh that's not what I understood :thinking: :sweat: Shouldn't/couldn't `ko` be able to rewrite those `knative-nightly` when releasing ?

adrcunha: It does rewrite then when building a new nightly release.

ImJasonH: We could have a separate `.ko.yaml.release` which we use when doing a real release, with `KO_CONFIG_PATH=.ko.yaml.release ko resolve ...`

PR #482, opened 2018-11-30
Support users to specify custom pod anntoations

### LINKED ISSUE #469: Allow users to specify pod anntoations
## Expected Behavior

Users can specify annotations for the underlying Pod created to execute the build.

## Actual Behavior

There is no such affordance.

## Additional Info

We should copy annotations on the requested build through to the underlying pod, so users can specify pod annotations.

mathspanda:
Fixes #469 

## Proposed Changes

  * Copy annotations on the build through to the underlying pod.

**Release Note**

```release-note
None
```


pkg/reconciler/build/resources/pod_test.go
```diff
@@ -468,6 +473,9 @@ func TestMakePod(t *testing.T) {
 				},
 				Spec: c.b,
 			}
+			if c.bAnnotations != nil {
```
shashwathi: I think you can update build annotation directly. Eg: 

```
b := &v1alpha1.Build{
	ObjectMeta: metav1.ObjectMeta{
		Name:        "build-name",
		Annotations: c.bAnnotations,
	},
	Spec: c.b,
}
```

pkg/reconciler/build/resources/pod.go
```diff
@@ -249,6 +250,15 @@ func makeCredentialInitializer(build *v1alpha1.Build, kubeclient kubernetes.Inte
 func MakePod(build *v1alpha1.Build, kubeclient kubernetes.Interface) (*corev1.Pod, error) {
 	build = build.DeepCopy()
 
+	// Copy annotations on the build through to the underlying pod to allow users
+	// to specify pod annotations.
+	annotations := map[string]string{
+		"sidecar.istio.io/inject": "false",
+	}
+	for key, val := range build.Annotations {
+		annotations[key] = val
```
shashwathi: If build has annotation `sidecar.istio.io/inject` set to `true` then previous annotation that sets to false is overridden. Is that desired behavior ?
If we are sure that `sidecar.istio.io/inject` should always be set to `false` then probably set this key after reading build annotations.  WDYT? @mathspanda @ImJasonH

pkg/reconciler/build/resources/pod_test.go
```diff
@@ -480,7 +488,15 @@ func TestMakePod(t *testing.T) {
 			}
 
 			if d := cmp.Diff(&got.Spec, c.want, ignorePrivateResourceFields); d != "" {
-				t.Errorf("Diff:\n%s", d)
+				t.Errorf("Diff spec:\n%s", d)
+			}
+
+			if c.bAnnotations != nil {
```
shashwathi: nit: This is just a suggestion

```
wantAnnotations := map[string]string{"sidecar.istio.io/inject": "false"}  // this indicates that pod annotation will always have this annotation set as default
if c.bAnnotations != nil {
	for key, val := range c.bAnnotations { // adding this loop suggests new set of keys will be added to default set
		wantAnnotations[key] = val 
	}
}
if d := cmp.Diff(got.Annotations, wantAnnotations); d != "" {
	t.Errorf("Diff annotations:\n%s", d)
}
```

I prefer this change because annotations will be checked on all test cases not just the one which has custom annotations set.

pkg/reconciler/build/resources/pod.go
mathspanda: Yeap. I think you are right. `sidecar.istio.io/inject` should be set last. Fixed.

mathspanda:
Fixed as you suggested. @shashwathi THX for review.

Issue #290, opened 2018-07-28
Reconcile pkg/{logs,logging}


mattmoor:
Build has both `./pkg/logging` and `./pkg/logs`, which will be confusing to new contributors.

PR #487, opened 2018-12-01
Add more vendored deps required for `dep-collector` 🤣

bobcatfish:
In #485 (#290 in build-pipeline)  I tried to fix the logic that automatically installs
`dep-collector` when it is not present on your system, but it looks like
I missed at least one dependency.

I think it is deeply ironic that we have to manually collect and add the
dependencies for a tool called `dep-collector` which is supposed to be
make it easy to managed deps 🤣

^^ Or am I doing something wrong?????

```release-note
NONE
```

Corresponds to https://github.com/knative/build-pipeline/pull/293

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -307,6 +307,14 @@
   version = "v2.0.1"
 
 [[projects]]
+  digest = "1:d917313f309bda80d27274d53985bc65651f81a5b66b820749ac7f8ef061fd04"
+  name = "github.com/sergi/go-diff"
+  packages = ["diffmatchpatch"]
+  pruneopts = "NUT"
+  revision = "1744e2970ca51c86172c8190fadad617561ed6e7"
+  version = "v1.0.0"
+
+[[projects]]
   digest = "1:e3707aeaccd2adc89eba6c062fec72116fe1fc1ba71097da85b4d8ae1668a675"
   name = "github.com/spf13/pflag"
   packages = ["."]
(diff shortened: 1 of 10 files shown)


mattmoor:
`go get` grabs HEAD, which has been notorious for breaking the Kubernetes folks (at least when outside of their control), and co-versioning via vendoring is a way to fix that.  That said, you don't want to add `dep-collector`'s dependencies, I think you'd want to add it to the list of `required = []` tools in `Gopkg.toml`.

adrcunha:
https://github.com/knative/build-pipeline/pull/296 should make things easier.

bobcatfish:
> That said, you don't want to add dep-collector's dependencies, I think you'd want to add it to the list of required = [] tools in Gopkg.toml.

@mattmoor will that make it grab the dependencies as well? and does that mean we can remove the dependencies from the .lock file as well - or does it just imply that `go dep` will manage those dependencies in the .lock file?

bobcatfish:
I think even with #489 we need this - I tried:
1. Deleting `licenseclassifier` and `go-diff` from the `Gopkg.lock` and from the `vendor` dir
2. Removing my `dep-collector` binary
3. Running `hack/update-deps.sh`

`go dep` didn't pick up the dependencies needed for `dep-collector` b/c I ended up with:

```
vendor/github.com/knative/test-infra/tools/dep-collector/licenses.go:28:2: cannot find package "github.com/google/licenseclassifier" in any of:
	/usr/local/google/home/christiewilson/Code/go/src/github.com/knative/build-pipeline/vendor/github.com/google/licenseclassifier (vendor tree)
	/usr/lib/google-golang/src/github.com/google/licenseclassifier (from $GOROOT)
	/usr/local/google/home/christiewilson/Code/go/src/github.com/google/licenseclassifier (from $GOPATH)
```

It seems like we still need to manually add the dependencies?

bobcatfish:
> @bobcatfish Adding it to required should be all you need, unless you need to pin specific versions.

well it didn't work for me @mattmoor but ill close this for now and reopen if god forbid this happens again

PR #503, opened 2018-12-15
check cluster Version 

zouyee:
## Proposed Changes
Knative requires a Kubernetes cluster v1.10 or newer. This makes it easier to
figure out if kubernetes meet the requirement.

**Release Note**

```
NONE

```
/lint


cmd/controller/main.go
```diff
@@ -79,6 +80,16 @@ func main() {
 	if err != nil {
 		logger.Fatalf("Error building kubernetes clientset: %v", err)
 	}
+	// Knative requires a Kubernetes cluster v1.10 or newer, making debugging hard. This makes it easier to
+	// figure out if kubernetes meet the requirement.
+	v, err := kubeClient.Discovery().ServerVersion()
```
ImJasonH: This seems like a useful thing to add to knative/pkg, so all the Knative components can reuse it.

mattmoor: Looks already pretty well factored, but other things could use this for sure.

shashwathi:
@zouyee :
Can you please explain why `apimachinery` package is copied over? 
- If your code requires newer version then please try to update library version instead of copying code over.

zouyee:
I am unable to determine the version relationship between k8s and knative, whether there is a correspondence and compatibility matrix doc?

PR #510, opened 2018-12-21
Add possibility for a client to cancel a build…

vdemeester:
… by updating the status of a build.

It will mark the build as failed and clean any pods related to it.

Related to https://github.com/knative/build-pipeline/issues/272
I tried that on `knative/build` first (as it's simpler :upside_down_face:)

cc @abayer @ImJasonH 

### DIFF
--- a/pkg/reconciler/build/build.go
+++ b/pkg/reconciler/build/build.go
@@ -153,6 +153,11 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
 		return nil
 	}
 
+	// If the build's status is cancelled, kill resources and update status
+	if isCancelled(build.Spec) {
+		return c.cancelBuild(build, logger)
+	}
+
 	// If the build hasn't started yet, validate it and create a Pod for it
 	// and record that pod's name in the build status.
 	var p *corev1.Pod
@@ -281,7 +286,33 @@ func (c *Reconciler) startPodForBuild(build *v1alpha1.Build) (*corev1.Pod, error
 	return c.kubeclientset.CoreV1().Pods(p.Namespace).Create(p)
 }
 
-// IsDone returns true if the build's status indicates the build is done.
+// isCancelled returns true if the build's spec indicates the build is cancelled.
+func isCancelled(buildSpec v1alpha1.BuildSpec) bool {
+	return buildSpec.Status == v1alpha1.BuildSpecStatusCancelled
+}
+
+func (c *Reconciler) cancelBuild(build *v1alpha1.Build, logger *zap.SugaredLogger) error {
+	logger.Warnf("Build has been cancelled: %v", build.Name)
+	build.Status.SetCondition(&duckv1alpha1.Condition{
+		Type:    v1alpha1.BuildSucceeded,
+		Status:  corev1.ConditionFalse,
+		Reason:  "BuildCancelled",
+		Message: fmt.Sprintf("Build %q was cancelled", build.Name),
+	})
+	if err := c.updateStatus(build); err != nil {
+		return err
+	}
+	if build.Status.Cluster == nil {
+		logger.Warnf("build %q has no pod running yet", build.Name)
+		return nil
+	}
+	if err := c.kubeclientset.CoreV1().Pods(build.Namespace).Delete(build.Status.Cluster.PodName, &metav1.DeleteOptions{}); err != nil {
+		return err
+	}
+	return nil
+}
+
+// isDone returns true if the build's status indicates the build is done.
 func isDone(status *v1alpha1.BuildStatus) bool {
 	cond := status.GetCondition(v1alpha1.BuildSucceeded)
 	return cond != nil && cond.Status != corev1.ConditionUnknown
--- a/pkg/reconciler/build/build_test.go
+++ b/pkg/reconciler/build/build_test.go
@@ -561,3 +561,62 @@ func TestTimeoutFlow(t *testing.T) {
 		t.Errorf("Unexpected build status %s", d)
 	}
 }
+
+func TestCancelledFlow(t *testing.T) {
+	b := newBuild("cancelled")
+
+	f := &fixture{
+		t:          t,
+		objects:    []runtime.Object{b},
+		client:     fake.NewSimpleClientset(b),
+		kubeclient: k8sfake.NewSimpleClientset(),
+	}
+	f.createServiceAccount()
+
+	r, i, k8sI := f.newReconciler()
+	f.updateIndex(i, b)
+	stopCh := make(chan struct{})
+	defer close(stopCh)
+	i.Start(stopCh)
+	k8sI.Start(stopCh)
+
+	ctx := context.Background()
+	if err := r.Reconcile(ctx, getKey(b, t)); err != nil {
+		t.Errorf("Not Expect error when syncing build")
+	}
+
+	buildClient := f.client.BuildV1alpha1().Builds(b.Namespace)
+	b, err := buildClient.Get(b.Name, metav1.GetOptions{})
+	if err != nil {
+		t.Errorf("error fetching build: %v", err)
+	}
+
+	// Get pod info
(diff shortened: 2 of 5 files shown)


vdemeester:
@ImJasonH 

> 1.) Is it normally common practice for a client to be able to update k8s resources' status this way? I'm thinking of something like Job, which runs to completion, but which if you want to cancel it you basically just delete it. In this case, we want to stop the build's execution without also deleting any record of that resource existing.

I really am not sure… I'm trying to look around for that :sweat_smile: 

> 2.) Once this is done we should also make the same change for TaskRun, and then PipelineRun, which should cancel (but not delete) its created TaskRuns. We might even just want to go straight to TaskRun and drop this PR. I'll leave that decision to up to you.

Yes, I did this here as it was simpler/quicker but I intend to follow-up on `build-pipeline`. I'm also fine doing it directly on `build-pipeline`, I just thought if we want it in `build-pipeline` we *may* want it there too :angel:

shashwathi:
@vdemeester Great PR

If build is completed then what is the point updating status to `Cancelled`?
Build is completed so all associated resources are deleted.
If build is running then `cluster` name would be set in build status. 
So I don't understand why there is a API call to get pod resource? and why not just call the delete API directly?

vdemeester:
> @vdemeester Great PR
> 
> If build is completed then what is the point updating status to `Cancelled`?
> Build is completed so all associated resources are deleted.

So, if you want to cancel the build (currently using the spec), then you want to mark it as `Cancelled` as soon as possible and stop the execution, thus deleting the pod real quick :angel:, same way we do on build timeout :upside_down_face:. (making me think there could be a small refactor to share code there).

> If build is running then `cluster` name would be set in build status.
> So I don't understand why there is a API call to get pod resource? and why not just call the delete API directly?

Indeed, there is no need.

abayer:
At Kubecon, @bobcatfish told me (in the context of https://github.com/knative/build-pipeline/pull/355) that I shouldn't be making changes to the status of one CRD while reconciling another one, so I jumped through some hoops to handle having the `PipelineRun`'s timeout apply to its `TaskRun`s. Not sure if that is a general thing where nothing should be modifying a CRD's status but its reconciler or if it's just that one CRD's reconciler shouldn't mess with another CRD's status.

shashwathi:
> So, if you want to cancel the build (currently using the spec), then you want to mark it as Cancelled as soon as possible and stop the execution, thus deleting the pod real quick 👼

I agree with this completely. I do not think you understood my question so let me rephrase
If build is completed (pods are not present and only build object exists), then what is the point of marking build as "Cancelled"?

I am thinking of this like a cycle.
build starts -> pod creation -> pod finished and deleted -> build completed. 

If user marks build as "Cancelled" after completion, I do not see how that has any effect on underlying resource(pod). I would expect in this case for status to be not changed at all. If build has finished then status should not be changed(IMO).

> same way we do on build timeout 🙃. (making me think there could be a small refactor to share code there).

Controller checks if build is still running and then cancels the build if timeout has passed. It doesn't cancel if build has completed.

vdemeester:
> I agree with this completely. I do not think you understood my question so let me rephrase
> If build is completed (pods are not present and only build object exists), then what is the point of marking build as "Cancelled"?
> 
> I am thinking of this like a cycle.
> build starts -> pod creation -> pod finished and deleted -> build completed.
> 
> If user marks build as "Cancelled" after completion, I do not see how that has any effect on underlying resource(pod). I would expect in this case for status to be not changed at all. If build has finished then status should not be changed(IMO).

Indeed there is no point, and if a user marks the build as `Cancelled` after completion, nothing should happen indeed. It should already be the case (https://github.com/knative/build/pull/510/files#diff-f2d9d6d36e3340295720d4f3e028cb14R152). If the build is completed (successfully or not), it won't ever go into the "cancel" block :angel: 

> Controller checks if build is still running and then cancels the build if timeout has passed. It doesn't cancel if build has already completed.

Yes, this is what happens.

test/e2e/simple_test.go
```diff
@@ -605,3 +605,47 @@ func TestSimpleBuildWithHybridSources(t *testing.T) {
...
+				Args:  []string{"sleep", "5000"}, // fails.
+			}},
+		},
+	}); err != nil {
+		t.Fatalf("Error creating build: %v", err)
+	}
+
+	// Wait for a little while
+	// FIXME(vdemeester) I would prefer something less flaky-prone
+	time.Sleep(20 * time.Second)
```
shashwathi: Probably consider adding polling function to check build status is "Running" for 10sec(? or more). That is better than wait solution IMO. WDYT @vdemeester ?

test/e2e/simple_test.go
```diff
@@ -605,3 +605,47 @@ func TestSimpleBuildWithHybridSources(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Error getting build %s: %v", buildName, err)
+	}
+	b.Spec.Status = v1alpha1.BuildSpecStatusCancelled
+	if _, err := clients.buildClient.builds.Update(b); err != nil {
+		t.Fatalf("Error update build status for %s: %v", buildName, err)
+	}
+
+	if _, err := clients.buildClient.watchBuild(buildName); err == nil {
+		t.Fatalf("watchBuild did not return expected `cancelled` error")
```
shashwathi: `watchBuild` function returns `build` object so you could consider verifying that build status is updated with expected reason("BuildCancelled") here.

pkg/reconciler/build/build_test.go
```diff
@@ -561,3 +561,62 @@ func TestTimeoutFlow(t *testing.T) {
...
+		t.Fatalf("build status did not specify podName: %v", b.Status.Cluster)
+	}
+	b.Spec.Status = v1alpha1.BuildSpecStatusCancelled
+
+	f.updateIndex(i, b)
+	if err := r.Reconcile(ctx, getKey(b, t)); err != nil {
+		t.Errorf("error syncing build: %v", err)
+	}
+
+	// Check that the build has the expected timeout status.
```
shashwathi: typo: I think you meant cancelled status here

pkg/reconciler/build/build.go
```diff
@@ -153,6 +153,36 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
+			Message: fmt.Sprintf("Build %q was cancelled", build.Name),
+		})
+		if err := c.updateStatus(build); err != nil {
+			return err
+		}
+		if build.Status.Cluster == nil {
+			logger.Warnf("build %q has no pod running yet", build.Name)
+			return nil
+		}
+		p, err := c.kubeclientset.CoreV1().Pods(build.Namespace).Get(build.Status.Cluster.PodName, metav1.GetOptions{})
```
shashwathi: I think I already mentioned this(?) in previous comment about reusing `build.Status.Cluster.PodName` for Delete API directly. Sorry to repeat it again

test/e2e/simple_test.go
vdemeester: @shashwathi yeah, that's why the FIXME :angel: not sure why I didn't use the polling functions here :sweat_smile: will update :wink:

pkg/reconciler/build/build.go
vdemeester: oh right, didn't update that :sweat:

shashwathi:
I noticed @ImJasonH had some comment earlier in thread so I am adding `hold` for him to add his review.

ImJasonH:
/lgtm
/approve
/hold cancel

Oh my goodness I didn't realize this was blocking on me, so so sorry to keep this held up so long. Thanks for this change! :heart:

vdemeester:
ah.. I may need to rebase and fix 😅

vdemeester:
Hum… the coverage build is a bit a pain :sweat_smile: as the threshold is higher than the current coverage on that file, any change trigger a failure… :sweat:

vdemeester:
| /test pull-knative-build-integration-tests
-- | --

PR #517, opened 2019-01-03
bump knative/caching

dprotaso:
## Proposed Changes

*  Includes the latest and greatest that `knative/caching` has to offer

**Release Note**

```release-note
NONE
```

### DIFF
--- a/hack/update-deps.sh
+++ b/hack/update-deps.sh
@@ -25,4 +25,7 @@ cd ${REPO_ROOT_DIR}
 # Ensure we have everything we need under vendor/
 dep ensure
 
+rm -rf $(find vendor/ -name 'OWNERS')
+rm -rf $(find vendor/ -name '*_test.go')
+
 update_licenses third_party/VENDOR-LICENSE "./cmd/*"
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -207,7 +207,7 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:826576a2dbea52eb350ed5c0fd3e53f1f24cd6bdcfaf89ff89239b8cd426cd0d"
+  digest = "1:ad0cca2383dc5678f6711c51cd38594906701e07d713998e3c9240e926f22867"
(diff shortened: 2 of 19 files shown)


hack/update-deps.sh
```diff
@@ -25,4 +25,7 @@ cd ${REPO_ROOT_DIR}
 # Ensure we have everything we need under vendor/
 dep ensure
 
+rm -rf $(find vendor/ -name 'OWNERS')
+rm -rf $(find vendor/ -name '*_test.go')
```
vdemeester: This shouldn't be required :thinking: (`dep` is able prunes tests in recent version — >= 0.4.x for sure)

dprotaso: Oddly enough it didn't work for me

```
dep:
 version     : v0.5.0
 build date  : 2018-07-26
 git hash    : 224a564
 go version  : go1.10.3
 go compiler : gc
 platform    : darwin/amd64
 features    : ImportDuringSolve=false
```

vdemeester: @dprotaso hum that is really weird then :thinking: :sweat:

dprotaso: Prune settings are off on a few dependent packages - that's why they appear

dprotaso: https://github.com/knative/build/blob/277d5bca5f217f4c34803e5b1d35ac2f98b228b7/Gopkg.toml#L58-L70

vdemeester: That said…

```
dep: WARNING: redundant prune option "go-tests" set for "k8s.io/code-generator"
dep: WARNING: redundant prune option "go-tests" set for "github.com/knative/caching"
dep: WARNING: redundant prune option "go-tests" set for "github.com/knative/test-infra"
```

So there should be no reason for `dep` to keep them…

adrcunha: @dprotaso for test-infra, `non-go` is false because we need the shell scripts. We don't care about tests. I agree that there must be a reason for dep to keep some tests, since by default they are pruned.

PR #519, opened 2019-01-07
update build with a dummy status first to avoid race condition of another watch event while the pod is being created

### LINKED ISSUE #518: duplicate build pods created intermittently
## Expected Behavior
A single pod is created for a given `build` resource

## Actual Behavior
Occasionally multiple pods are created suggesting a timing issue.

Unfortunately it's not easily reproducible however we found the issue when upgrading Jenkins X from knative build `0.1.x` to `0.2.x`.  

## Additional Info
After noodling the code my best guess was there's a timing issue where a build resource is created and a pod is started https://github.com/knative/build/blob/d2a0ec1/pkg/reconciler/build/build.go#L173 meanwhile there's an update to the build resource which results in another `reconcile` being called, as the [build status is nil](https://github.com/knative/build/blob/d2a0ec1/pkg/reconciler/build/build.go#L159) another pod is started.  
I have a PR on the way with a suggested fix which updates a build status before starting the pod, I've deployed this to our own Jenkins X infra and not seen a duplicate pod since so I'm pretty

rawlingsj:
fixes #518

### DIFF
--- a/pkg/reconciler/build/build.go
+++ b/pkg/reconciler/build/build.go
@@ -170,6 +170,30 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
 			}
 			return err
 		}
+
+		// Add a unique suffix to avoid confusion when a build
+		// is deleted and re-created with the same name.
+		// We don't use GenerateName here because k8s fakes don't support it.
+		podName, err := resources.GetUniquePodName(build.Name)
+		if err != nil {
+			return err
+		}
+
+		// update with a dummy status first to avoid race condition of another event while the pod is being created
+		build.Status = v1alpha1.BuildStatus{
+			Builder: v1alpha1.ClusterBuildProvider,
+			Cluster: &v1alpha1.ClusterSpec{
+				Namespace: build.Namespace,
+				PodName:   podName,
+			},
+			StartTime: &metav1.Time{
+				Time: time.Now(),
+			},
+		}
+		if err := c.updateStatus(build); err != nil {
+			return err
+		}
+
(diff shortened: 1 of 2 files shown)


pkg/reconciler/build/build.go
```diff
@@ -170,6 +170,31 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
+			Builder: v1alpha1.ClusterBuildProvider,
+			Cluster: &v1alpha1.ClusterSpec{
+				Namespace: build.Namespace,
+				PodName:   podName,
+			},
+			StartTime: &metav1.Time{
+				Time: time.Now(),
+			},
+		}
+		err = c.updateStatus(build)
```
shashwathi: `c.updateStatus(build)` is called twice.

rawlingsj: doh thanks - removed the extra line

ImJasonH:
Thanks for finding and fixing this! I wonder if there's some way to test this more conclusively, preferrably with an integration test but I'd even take a real e2e test if it's the only way.

It's a fiddly bug that's dependent on a lot of stuff outside of our control, so I'd really like to be more sure that it's not going to crop back up.

scothis:
I was experiencing the same issue with duplicate build pods. Applying this PR resolved the issue as best as I can tell. Thanks @rawlingsj

rawlingsj:
> I was experiencing the same issue with duplicate build pods. Applying this PR resolved the issue as best as I can tell. 

Great - glad it helped!

@shashwathi thanks for merging :)

PR #530, opened 2019-01-11
Use the default build/unit/integration test runners

adrcunha:
Less boilerplate code, more checks.

Requires updating test-infra.


adrcunha:
/hold

And now my client has gone crazy. Ugh.

adrcunha:
My branch is busted, recreating in 531.

PR #543, opened 2019-01-28
Resync every 10h instead of every 30s

### LINKED ISSUE #541: Controller stops processing builds after ~400 builds
/kind bug

## Expected Behavior
I expect to be able to run an arbitrary number of builds to my cluster without cleaning up my builds.

## Actual Behavior
After ~400-500 builds, Knative stops processing new builds. It doesn't look like the controller is falling over, and it doesn't log any errors. The steps below create many builds quickly, but I initially observed this problem as my cluster accumulated builds over many days. 

Deleting old builds fixes the problem - the controller starts processing builds again.

```
REGION=us-west1
CLUSTER_NAME=$USER-knative-build

gcloud beta container clusters create $CLUSTER_NAME \
    --region=$REGION \
    --cluster-version=latest \
    --machine-type=n1-standard-16 \
    --enable-autoscaling --min-nodes=1 --max-nodes=3 \
    --enable-autorepair \
    --no-enable-basic-auth \
    --no-issue-client-certificate \
    --enable-network-policy

gcloud container clusters get-credentials $CLUSTER_NAME --region=$REGION

kubectl create clusterrolebinding cluster-admin-binding \
    --clusterrole=cluster-admin \
    --user=$(gcloud config get-value core/account)

kubectl apply -f https://github.com/knative/build/releases/download/v0.3.0/release.yaml

for i in $(seq 1 500); do
sed -e "s/{}/$i/" <<EOF | kubectl apply -f -
apiVersion: build.knative.dev/v1alpha1
kind: Build
metadata:
  name: sleep-{}
spec:
  steps:
    - image: busybox
      args: ["sleep", "10"]
EOF
done
```

ImJasonH:
Fixes #541

## Proposed Changes

* Resync our custom resources every 10 hours, and rely on build/pod watch updates to trigger syncs in most cases

Resyncing every 30s seems to lead us to hit rate limits that we never recover from when there are too many builds in the system at a given time. Serving also switched to resyncing every 10h for similar scaling reasons.

**Release Note**

```release-note
NONE
```

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -44,6 +44,7 @@ import (
 const (
 	threadsPerController = 2
 	logLevelKey          = "controller"
+	resyncPeriod         = 10 * time.Hour
 )
 
 var (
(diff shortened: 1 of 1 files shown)


ImJasonH:
```
$ kubectl get build sleep-500 -o=custom-columns-file=./test/columns.txt
NAME        TYPE        STATUS   CREATED                START                  END
sleep-500   Succeeded   True     2019-01-25T21:34:34Z   2019-01-25T21:39:58Z   <none>
```

(End time is not populated for any of the builds created in the load test, likely another bug)

sebgoa:
We ran into this at ~200 builds. Happy to test a fix.

PR #544, opened 2019-01-29
dep update of test-infra repository

nbarthwal:


adrcunha:
@nbarthwal This will break `hack/release.sh`, please update it in this PR. Just follow what we have in serving and eventing.

PR #549, opened 2019-02-02
complete boilerplate file

zouyee:
Fixes #

## Proposed Changes

* complete boilerplate file 

**Release Note**

```
NONE

```

### DIFF
--- a/hack/boilerplate/boilerplate.sh.txt
+++ b/hack/boilerplate/boilerplate.sh.txt
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+
+# Copyright 2019 The Knative Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


hack/boilerplate/boilerplate.sh.txt
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+# Copyright 2019 The Knative Authors
```
vdemeester: Shouldn't it be `2018-2019`

adrcunha:
> This seems like something that should be shared in knative/pkg (or knative/test-infra?) and vendored into this repo, since presumably it will be used by all the Knative repos. I don't want Build to be the only repo that enforces/automates this while the other repos do something else.

This is a copy of the script in serving. It's a convenience script, but I'm not sure how much usage it gets. If you think it's just adding unnecessary stuff to build, feel free to reject it.

/uncc @adrcunha

PR #552, opened 2019-02-07
replace map[string]x with string set

vagababov:
For: https://github.com/knative/serving/issues/3049

## Proposed Changes

* replace maps with sets

/lint


mattmoor:
LOL
/lgtm

Issue #542, opened 2019-01-25
Knative build creates duplicate pods


meteatamel:
Knative build consistently creates duplicate builds. This is my build definition:

```yaml
apiVersion: build.knative.dev/v1alpha1
kind: Build
metadata:
  name: build-helloworld-csharp-gcr
spec:
  source:
    git:
      url: "https://github.com/knative/docs.git"
      revision: "v0.1.x"
    subPath: "serving/samples/helloworld-csharp/"
  steps:
  - name: build-and-push
    image: "gcr.io/kaniko-project/executor:v0.6.0"
    args:
    - "--dockerfile=/workspace/Dockerfile"
    - "--destination=gcr.io/MY_GCP_PROJECT/helloworld-csharp:v1"`
```

```bash
> kubectl get build
No resources found.
> kubectl apply -f build-helloworld-csharp-gcr.yaml 
build.build.knative.dev "build-helloworld-csharp-gcr" created
> kubectl get build
NAME                          AGE
build-helloworld-csharp-gcr   5s
> kubectl get pods
NAME                                              READY     STATUS     RESTARTS   AGE
build-helloworld-csharp-gcr-pod-e7d4b1            0/1       Init:2/3   0          8s
build-helloworld-csharp-gcr-pod-ea29fd            0/1       Init:1/3   0          7s
```
The first one gets stuck in this step:
```bash
INFO[0083] Taking snapshot of full filesystem...
```

The second one builds and pushes to GCR.

sebgoa:
@ImJasonH we seem to face this issue as well in 0.3.0

utka:
I see the same bug for my project

ImJasonH:
I'm able to reproduce this, in about 2/11 tries:

```
$ while true; do kubectl apply -f build.yaml; sleep 10; kubectl get build build -ojsonpath="{.status.cluster.podName}"; echo; kubectl get pods; kubectl delete builds --all; done
build.build.knative.dev/build created
build-pod-3d7578
NAME               READY   STATUS     RESTARTS   AGE
build-pod-3d7578   0/1     Init:2/3   0          13s
build.build.knative.dev "build" deleted
build.build.knative.dev/build created
build-pod-0633e0
NAME               READY   STATUS     RESTARTS   AGE
build-pod-0633e0   0/1     Init:2/3   0          12s
build-pod-58bfde   0/1     Init:2/3   0          13s  # <---- BOOM
build.build.knative.dev "build" deleted
build.build.knative.dev/build created
[...]

PR #555, opened 2019-02-12
"BuildStarted" status

tzununbekov:
Dummy pod name from #519 appears in pod Get response while `startPodForBuild` is being processed (which may take few seconds in rare cases).
Fixes https://github.com/knative/build/issues/542#issuecomment-459802297

## Proposed Changes

*   `BuildStarted` condition instead of dummy status and non-existent pod name 

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/build/build.go
+++ b/pkg/reconciler/build/build.go
@@ -214,13 +214,6 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
 
 	// Update the build's status based on the pod's status.
 	build.Status = resources.BuildStatusFromPod(p, build.Spec)
-
-	// Check if build has timed out; if it is, this will set its status
-	// accordingly.
-	if err := c.checkTimeout(build); err != nil {
-		return err
-	}
-
 	return c.updateStatus(build)
 }
 
@@ -228,6 +221,10 @@ func (c *Reconciler) updateStatus(u *v1alpha1.Build) error {
 	newb, err := c.buildclientset.BuildV1alpha1().Builds(u.Namespace).Get(u.Name, metav1.GetOptions{})
 	if err != nil {
 		return err
+	}
+	cond := u.Status.GetCondition(v1alpha1.BuildSucceeded)
+	if cond != nil && cond.Status == corev1.ConditionFalse {
+		return fmt.Errorf("build %q is failed, can't update status", u.Name)
 	}
 	newb.Status = u.Status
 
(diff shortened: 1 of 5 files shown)


pkg/reconciler/build/build.go
```diff
@@ -158,10 +158,23 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
+	if isStarted(&build.Status) {
+		return nil
+	}
+
 	// If the build hasn't started yet, validate it and create a Pod for it
 	// and record that pod's name in the build status.
 	var p *corev1.Pod
-	if build.Status.Cluster == nil || build.Status.Cluster.PodName == "" {
+	if build.Status.StartTime == nil {
+		build.Status.SetCondition(&duckv1alpha1.Condition{
```
ImJasonH: Would it be sufficient to set the `StartTime` here and update the status immediately? AIUI we basically just need to ensure this branch isn't taken for the same build twice.

I think we could also accomplish that by generating a `podName` and persisting it immediately (then checking if `podName != ""`), but users would sometimes see a pod name that doesn't exist yet, which could be confusing.

ImJasonH:
cc @rawlingsj as author of #519

pkg/reconciler/build/build.go
shashwathi: > which could be confusing.

True but the build is still in "Unknown" state so it doesn't seem that bad.

tzununbekov: > Would it be sufficient to set the `StartTime` here and update the status immediately?
that may work as well. If we agree that dummy pod name should be removed, I can try this

> True but the build is still in "Unknown" state so it doesn't seem that bad.
I think it makes build statuses very unreliable - `Unknown` means that we can't trust build pod name that we get from API and `succeeded` means that build already passed. What if we need to "watch" build pod while build is not succeeded?

pkg/reconciler/build/build.go
```diff
@@ -235,10 +226,16 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
 }
 
 func (c *Reconciler) updateStatus(u *v1alpha1.Build) error {
+	buildStatus.Lock()
```
ImJasonH: I'd like to avoid this global lock across all build status updates. Can we get away with locking around each build that needs to be updated instead? e.g., Use a `map[string]sync.Mutex` where the key is the build key (namespace+name)?

pkg/reconciler/build/timeouts_handler.go
```diff
@@ -0,0 +1,81 @@
...
+func (t *TimeoutSet) checkPodTimeouts(namespace string) error {
+	builds, err := t.buildclientset.BuildV1alpha1().Builds(namespace).List(metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+	for _, build := range builds.Items {
+		if isDone(&build.Status) {
+			continue
+		}
+		r := Reconciler{
```
ImJasonH: I don't really like having two `Reconciler` instances involved at the same time, I think we'll get confused about which one is responsible for what.

How about we expose `TimeoutSet.CheckTimeout(build)` and inside the `Reconcile` method (which is called on every pod event) we check whether the build has timed out since the last event.

Separate from that, a `TimeoutSet` is also running goroutines which wait in the background for the timeout to expire and update builds, but just in case those don't happen we'll at least be able to enforce the timeout on the next pod event.

pkg/reconciler/build/timeouts_handler.go
```diff
@@ -0,0 +1,81 @@
...
+// TimeoutHandler walks through all namespaces, gets list of builds and checks timeouts
+func (t *TimeoutSet) TimeoutHandler() error {
+	ticker := time.NewTicker(checkInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-t.stopCh:
+			return nil
+		case <-ticker.C:
```
ImJasonH: Rather than polling for timeouts, what do you think about having a goroutine per ongoing build that does:

```go
select {
  case <-t.stopCh:
    // controller stopping, give up
  case <- build.finished:
    // build finished, give up
  case <- time.After(timeoutDur):
    // timeout elapsed
    t.stopBuild(build)
}
```

This should stop builds ~immediately upon timing out, which is nice, but also avoids unnecessary polling when we can tell the timeout is still minutes (or hours!) away.

We'd also need to spawn these goroutines when the controller starts too, in case the controller dies and restarts, to make sure ongoing builds have their timeouts enforced.

tzununbekov:
> I'm a bit worried this PR has started to solve three problems:
> 
> 1. increasing the resync time
> 2. enforcing timeouts
> 3. handling duplicate pod creation cleaner
> 
> For now the PR is fairly small and easy to follow, but I'm worried about how much bigger it will get when there are tests. If it's possible to split this into three PRs that might be easier to review.

@ImJasonH thanks for review, I'm totally agree with it. Going to close this PR and open new ones with all comments taken into consideration.

PR #581, opened 2019-03-20
Update to k8s 1.12.6 and pkg @ HEAD

mattmoor:
I had to pin a ton of packages to get things working.  In every case I basically just pinned the version in knative/serving.


mattmoor:
Ran `./hack/update-codegen.sh` with the new code-generator

mattmoor:
/hold

I may have screwed up the Status inlining.

mattmoor:
Nevermind, it was the `Status` field in spec that I was seeing.
/hold cancel

Gopkg.toml
```diff
@@ -19,36 +19,70 @@ required = [
...
+  revision = "511bab8e55decfbe784cd2d1868fa5cda62dc6c1"
+
+[[override]]
+  name = "github.com/census-instrumentation/opencensus-proto"
+  # What serving has in Gopkg.lock
+  revision = "7f2434bc10da710debe5c4315ed6d4df454b4024"
+
+[[override]]
+  name = "golang.org/x/oauth2"
+  # What serving has in Gopkg.lock
```
ImJasonH: This comment will get out of date almost immediately. Is there a better way to describe why these overrides are here? Why do we need them?

mattmoor: Now that these are locked, we can actually just remove these...  🙄

PR #588, opened 2019-03-29
Add a label knative.dev/crd-install to CRDs to allow installing CRDs in a separate pass.

evankanderson:
## Proposed Changes

- Label CRDs so that they can be installed first with `kubectl -l knative.dev/crd-install=true YAML_FILE`

**Release Note**

```release-note
- CRDs are now labelled with knative.dev/crd-install=true to allow installing without race conditions or kubectl errors.
```


ImJasonH:
Is there a corresponding docs change to describe how to use these labels when installing?

evankanderson:
There will be a docs change, which may need to be back-integrated into release-0.5

PR #590, opened 2019-04-02
Move to the v1beta1 Conditions implementation.

mattmoor:
See here for more context on what's changed (no breaking changes): https://github.com/knative/pkg/pull/361


vdemeester:
@mattmoor `knative/pkg` now uses `apps` instead of `extensions` for `Deployment` (see https://github.com/knative/pkg/commit/8b3dc0d76d96f5ae9711e3e503f37c4c148129fc). Might be why the e2e fails

PR #594, opened 2019-04-10
add ide ignore files

zouyee:
## Proposed Changes

* add ide ignore files

**Release Note**

```NONE

```


.gitignore
```diff
@@ -15,6 +15,7 @@
 
 # Fortran module files
 *.mod
+*.sum
```
vdemeester: Why `*.sum` ?
If we ever switch to `Go` modules `*.sum` and `*.mod` we will need to add them back

zouyee: Currently we have not modified the vendor tool to `go mod`, so we need to exclude `go.sum` and `go.mod`, also, when somebody using IDE, the above two files may be automatically generated.

ImJasonH:

PR #595, opened 2019-04-10
Add timeout to kubectl delete to avoid later failures

chaodaiG:
Fixes https://github.com/knative/test-infra/issues/676

## Proposed Changes

By default, `kubectl delete` estimates a timeout if not provided, which makes the deletion not always guaranteed. Add timeout explicitly to avoid later failures


chaodaiG:
@srinivashegde86 , an approve label is missing

PR #596, opened 2019-04-11
git-init update and fixes

vdemeester:
## Proposed Changes

- check for errors or at least log them trim spaces from url : if the
- given url contains some spaces (like non-breaking spaces),
  `git-init` may fail with `protocol ' https' is not supported`. This
  fixes that.

Related to https://github.com/tektoncd/pipeline/pull/677 and https://github.com/tektoncd/pipeline/pull/750

### DIFF
--- a/cmd/git-init/main.go
+++ b/cmd/git-init/main.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 
 	"github.com/knative/pkg/logging"
 	"go.uber.org/zap"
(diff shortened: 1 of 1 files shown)


cmd/git-init/main.go
```diff
@@ -87,16 +91,18 @@ func main() {
 		run(logger, "git", "init")
```
chmouel: Why you do that twice?

And perhaps check if path+"./git" exist already ?

chmouel:
Perhaps add this review too https://github.com/tektoncd/pipeline/pull/747

zouyee:
I think `path` also need to be validated.

cmd/git-init/main.go
vdemeester: @chmouel it's not on the same "branch", so still needed

cmd/git-init/main.go
```diff
@@ -87,16 +91,18 @@ func main() {
...
 	}
 
-	run(logger, "git", "remote", "add", "origin", *url)
-	err = runOrFail(logger, "git", "fetch", "--depth=1", "--recurse-submodules=yes", "origin", *revision)
-	if err != nil {
+	trimedURL := strings.TrimSpace(*url)
+	runOrFail(logger, "git", "remote", "add", "origin", trimedURL)
+	if err = run(logger, "git", "fetch", "--depth=1", "--recurse-submodules=yes", "origin", *revision); err != nil {
```
zouyee: Could we consider using context to handler command hangs? Today we encountered a situation which the command was stuck, and there was a problem with the underlying storage.

vdemeester: @zouyee we may, with a default `timeout`, question is then, what would be a "sane" timeout ?

ImJasonH:

Issue #599, opened 2019-04-17
Build reported as a timeout when it actually succeeded


scothis:
## Expected Behavior

Successful builds completed before the timeout are reported as successful.

## Actual Behavior

Successful builds can be reported as a timeout.

In this case, the build and pod are created at 2019-04-17T21:02:21Z, and finish 31 seconds later at 2019-04-17T21:02:52Z. Instead of the build's `Succeeded` condition reporting `True` the message is `Build "square-build" failed to finish within "10m0s"`.

build
```
apiVersion: build.knative.dev/v1alpha1
kind: Build
metadata:
  creationTimestamp: "2019-04-17T21:02:21Z"
  generation: 1
  labels:
    projectriff.io/function: square
  name: square-build
  namespace: default
  resourceVersion: "2531684"
  selfLink: /apis/build.knative.dev/v1alpha1/namespaces/default/builds/square-build
  uid: 17a0e49f-6154-11e9-9a6b-42010a80016c
spec:
  Status: ""
  serviceAccountName: riff-build
  source:
    git:
      revision: master
      url: https://github.com/projectriff/fats.git
    subPath: functions/uppercase/node
  template:
    arguments:
    - name: IMAGE
      value: gcr.io/cf-sandbox-sandrews/square
    - name: CACHE
      value: cache
    kind: ClusterBuildTemplate
    name: riff-cnb
  timeout: 10m0s
  volumes:
  - name: cache
    persistentVolumeClaim:
      claimName: build-cache-square
status:
  builder: Cluster
  cluster:
    namespace: default
    podName: square-build-pod-4cad55
  completionTime: "2019-04-17T21:02:54Z"
  conditions:
  - lastTransitionTime: "2019-04-17T21:02:54Z"
    message: Build "square-build" failed to finish within "10m0s"
    reason: BuildTimeout
    status: "False"
    type: Succeeded
  startTime: "2019-04-17T20:52:54Z"
  stepsCompleted: null
```

pod (sorry there are so many steps)
```
apiVersion: v1
kind: Pod
metadata:
  annotations:
    kubernetes.io/limit-ranger: 'LimitRanger plugin set: cpu request for container
      nop; cpu request for init container build-step-credential-initializer; cpu request
      for init container build-step-git-source-0; cpu request for init container build-step-prepare;
      cpu request for init container build-step-detect; cpu request for init container
      build-step-analyze; cpu request for init container build-step-build; cpu request
      for init container build-step-export'
[...]

PR #600, opened 2019-04-25
fix build uncorrect status

zouyee:
Fixes https://github.com/knative/build/issues/599

## Proposed Changes

* check build status before stop it.

**Release Note**

```NONE

```


zouyee:
@mattmoor  PTAL

ImJasonH:
Hi @zouyee -- sorry it's taken so long to get to this PR 😅 

If this bug is still outstanding and reproducible, please add a test case that faiils without your code change to prevent accidentally regressing in some future refactor. A unit test change would also be nice.

ImJasonH:

PR #611, opened 2019-05-27
Move build onto the shared controller.HandleAll

mattmoor:
See also: knative/pkg#428

/hold

Until upstream change lands.


ImJasonH:
Superceded by #616

